PR #6, opened 2018-10-26
Make sure Kotlin output is ktlint-compliant

mdboom:
This makes sure that the Kotlin we're generating is `ktlint` compliant, and adds a test to run the output against `ktlint` if it's available.


Dexterp37:
@mdboom looks like this is failing to execute on CI, could you please check this out?

mdboom:
This is passing now.

PR #7, opened 2018-10-28
Update kotlin template to follow ktlint rules

jonalmeida:
I've started a thread at https://github.com/mozilla-mobile/android-components/issues/1193 where we can consider this approach or just not run ktlint on `build/` directories.

 ¯\\\_(ツ)\_/¯


jonalmeida:
Closing because I didn't see #6 before making this. 😅

PR #14, opened 2018-11-13
Enforce that event lifetime is ping

mdboom:
As discussed here: https://github.com/mozilla-mobile/android-components/pull/1308#discussion_r233093823

The event metric should always have a lifetime of ping.

### DIFF
--- a/glean_parser/schemas/metrics.1-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.1-0-0.schema.yaml
@@ -1,4 +1,4 @@
-$schema: http://json-schema.org/draft-06/schema#
+$schema: http://json-schema.org/draft-07/schema#
 title: Metrics
 description: |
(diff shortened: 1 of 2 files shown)


glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -1,4 +1,4 @@
-$schema: http://json-schema.org/draft-06/schema#
+$schema: http://json-schema.org/draft-07/schema#
```
Dexterp37: Is this new schema version supported on the pipeline side?

mdboom: It doesn't need to be -- this is only the schema for the `metrics.yaml` file, which is validated in this library and where we have control over the version of `jsonschema` used.

Dexterp37: Good, thanks for clarifying!

PR #15, opened 2018-11-19
Fix 1504200: Add ping validation feature to glean_parser

mdboom:
This adds a feature to glean_parser to validate pings.  It's basically a straightforward use of the `jsonschema` library that the backend uses to do that.

I've just copied the ping schema into this repo rather than doing anything fancy like fetching it from a URL because:

1. We don't want to fetch from the URL every time we run a unit test in Glean.  We could cache it, but cache invalidation is hard :cry:
2. We're already relying on changes to the ping schema that aren't yet merged into `mozilla-pipeline-schemas`.  At least until things settle down we probably want to be more nimble and have the ability to put what we need to in here.

We definitely can and should revisit that later.

### DIFF
--- a/glean_parser/validate_ping.py
+++ b/glean_parser/validate_ping.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+Validates the contents of a glean ping against the schema.
+"""
+
+import functools
+import io
+import json
+from pathlib import Path
+import sys
+
+import jsonschema
+
+from . import util
+
+
+ROOT_DIR = Path(__file__).parent
+SCHEMAS_DIR = ROOT_DIR / 'schemas'
+
+
+PING_SCHEMA_DEFAULT_URL = (
+    'https://raw.githubusercontent.com/mozilla-services/'
+    'mozilla-pipeline-schemas/dev/schemas/glean/baseline/'
+    'baseline.1.schema.json'
+)
+
+
+@functools.lru_cache(maxsize=1)
+def _get_ping_schema(schema_url):
+    contents = util.fetch_remote_url(
+        schema_url,
+        cache=(schema_url != PING_SCHEMA_DEFAULT_URL)
+    )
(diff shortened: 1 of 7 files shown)


glean_parser/schemas/baseline.1.schema.json
```diff
@@ -0,0 +1,424 @@
+{
```
Dexterp37: I think I have strong feelings about this, given that I messed up quite a few time Desktop Telemetry due to out of sync dependencies (that *I* forgot to update :) ).

What about we keep this generic, drop the schema off this repo, pass in a path to the schema when invoking the command?

tests/test_validate_ping.py
```diff
@@ -0,0 +1,41 @@
...
+# http://creativecommons.org/publicdomain/zero/1.0/
+
+import io
+import json
+
+from glean_parser import validate_ping
+
+
+def test_validate_ping():
+    content = {
```
Dexterp37: If we're moving the schema off this repo, let's keep this shorter: let's use a small JSON and provide a sample schema stub.

glean_parser/validate_ping.py
```diff
@@ -0,0 +1,50 @@
...
+ROOT_DIR = Path(__file__).parent
+SCHEMAS_DIR = ROOT_DIR / 'schemas'
+
+
+@functools.lru_cache(maxsize=1)
+def _get_ping_schema():
+    return util.load_yaml_or_json(SCHEMAS_DIR / 'baseline.1.schema.json')
+
+
+def validate_ping(ins, outs):
```
Dexterp37: We could probably ingest a file path here as well, right?

mdboom:
With respect to managing the schema:

- Since this is for unit testing, we shouldn't just link to the current HEAD, since all historical versions of glean wouldn't be expected to always work with the latest version of the schema, and this will make things like `git blame` much harder.  At a minimum we should link to a particular git revision of the schema.  Down the road when things settle down, we could link to a particular version number of the schema, but the overhead of managing version numbers in `mozilla-pipeline-schemas` is too high for the current phase of the project.

- If we link to a specific revision, managing the cache should also be much easier.

mdboom:
I've updated this to allow the end user to specify the schema URL.  It is cached, unless it's pointing to master which is a moving target and probably shouldn't be cached.

Thanks for another round of review, @Dexterp37

PR #16, opened 2018-11-19
1507792: Better error message when metrics.yaml is empty

mdboom:
https://bugzilla.mozilla.org/show_bug.cgi?id=1507792

@georgf

### DIFF
--- a/glean_parser/parser.py
+++ b/glean_parser/parser.py
@@ -98,6 +98,10 @@ def _load_metrics_file(filepath):
         yield f"{filepath}: {str(e)}"
         return {}
 
+    if metrics_content is None:
+        yield f"{filepath}: metrics file can not be empty."
+        return {}
+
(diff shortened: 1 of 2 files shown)


glean_parser/parser.py
```diff
@@ -58,6 +58,10 @@ def _load_metrics_file(filepath):
         yield f"{filepath}: {str(e)}"
         return {}
 
+    if metrics_content is None:
+        yield f"{filepath}: No JSON content found."
```
georgf: Nit: YAML content vs. JSON content.

I'd change the message to be more about what we want from the user (instead of the technical requirement).
Maybe something like:
> Metrics file can't be empty. Please supply a valid YAML file for: {filepath}

mdboom: Sure thing.  Done.

PR #18, opened 2018-11-20
1507788: Support . in category names

mdboom:


glean_parser/util.py
```diff
@@ -58,6 +58,22 @@ def ensure_list(value):
     return value
 
 
+def camelize(value):
```
Dexterp37: non blocking nit: these two functions only differ by the last argument passed to `inflection.camelize`. Do we really need 2 functions? If yes, can you mention their difference in their docstrings?

mdboom: These are used as filters in jinja2 templates, so for convenience and readability there, they have only one argument.  That second argument controls whether the first character in the output is upper case.

georgf:
Can you summarize the change?
That would help with diving into the review.

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -14,6 +14,10 @@ definitions:
     type: string
     pattern: "^[a-z_][a-z0-9_]*$"
 
+  dotted_snake_case:
+    type: string
+    pattern: "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})*$"
```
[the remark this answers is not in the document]
mdboom: I think we should determine that.  It was 30 characters, but if we're going to use it for heirarchy, it should probably be extended.  Maybe 50?  We should be conservative because it will ultimately affect ping sizes.

PR #22, opened 2018-12-13
Add a TimeUnit enum for time based metrics

Dexterp37:
Timespan and TimingDistribution both require a time_unit parameter, which is an enum users can
pick the resolution from. This adds validation for the possible values and enforces the correct code
generation.

### DIFF
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -144,9 +144,19 @@ class Counter(Metric):
     typename = 'counter'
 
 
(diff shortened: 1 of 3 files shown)


mdboom:
I guess this will need to be tagged after merge and the version number updated on the android-components side...

Dexterp37:
> I guess this will need to be tagged after merge and the version number updated on the android-components side...

Right! Let me merge this and tag a release. I'll bump the version number in the PR for the timespan.

PR #23, opened 2018-12-13
Bump version 0.9.0 -> 0.10.0

Dexterp37:


mdboom:
Thanks!  As documented [here](https://github.com/mozilla/glean_parser/blob/master/CONTRIBUTING.rst#deploying), if you use the `bumpversion` command, it also will create the tag automatically (which I always have trouble getting right, personally).  We should probably also start updating `HISTORY.rst` as described there, but I haven't been so far, and I don't think it's critical yet while we're still in 0.x.x.

In any case, I'm going to close this and do that.

mdboom:
No problem!

PR #24, opened 2018-12-14
1513617: Turn metrics.yaml schema into formal docs

mdboom:
The format of the `metrics.yaml` was already documented, just in a not-too-user-ffriendly form in the associated JSON schema file.

This uses that content to create a Sphinx page in the docs here, and also makes a few other organizational changes to the docs.

This also adds a Travis deployment so that the docs will automatically build and deploy every time we tag a new release.

You can preview the build of the docs here: https://mozilla.github.io/glean_parser/ (or you can check out the repo, install the dev dependencies and type `make docs`).

(Eventually, when we have a single repo for glean things, this documentation will unify the parser side and API side, but for now, I think it makes sense to keep this here with the parser where the schema lives.)


glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -111,8 +130,8 @@ definitions:
         title: Send in pings
         description: |
           Which pings the metric should be sent on. If not specified, the metric
-          is sent on the "default ping", which is the "events" ping for events,
-          and the "metrics" ping for everything else. Most metrics don't need to
+          is sent on the "default ping", which is the `events` ping for events,
```
Dexterp37: nit: let's also mention the `default` keyword here, used to represent default pings

Dexterp37:
> 
> 
> > Any chance you could also update glean's readme to point to these docs?
> 
> This PR adds:
> 
> ```
> Documentation
> -------------
> 
> The full documentation is available `here <https://mozilla.github.io/glean_parser/>`__.
> ```
> 
> Also, I updated the link in the banner on the github project to point to this.

Sorry, that was me being confusing. I meant: can you add a link to [this](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/README.md) ? :D

PR #30, opened 2019-02-07
Explicitly test for slash in names, categories and labels

mdboom:
As mentioned [here](https://github.com/mozilla-mobile/android-components/pull/1984#discussion_r254790281) we should be extra careful that we don't get slashes into names, categories or labels.  This adds test for that at build time (it was already being checked for, but there weren't explicit tests for it).

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -214,3 +214,58 @@ def test_parser_reserved():
     all_metrics = parser.parse_metrics(contents, {'allow_reserved': True})
     errors = list(all_metrics)
     assert len(errors) == 0
+
+
+def test_parser_slash_in_name():
+    contents = [
+        {
+            'glean.baseline': {
+                'metric/with_slash': {
+                    'type': 'string',
+                },
+            },
+        },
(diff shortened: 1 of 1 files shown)


tests/test_parser.py
```diff
@@ -214,3 +214,58 @@ def test_parser_reserved():
     all_metrics = parser.parse_metrics(contents, {'allow_reserved': True})
     errors = list(all_metrics)
     assert len(errors) == 0
+
+
+def test_parser_slash_in_name():
```
Dexterp37: This looks great! I wonder if we should do the same for the `#` character, which we're using as a separator in glean [here](https://github.com/mozilla-mobile/android-components/blob/b83fae2c1b1a444b996d31a5b05d1f73737a004e/components/service/glean/src/main/java/mozilla/components/service/glean/storages/GenericScalarStorageEngine.kt#L115). What do you think?

mdboom: That's a good idea.  I've updated this to test for more cases like this.

PR #39, opened 2019-03-12
1528260: Simplify and implement metric expiry

mdboom:
The proposal isn't quite finished, but this is a first pass at an implementation anyway, so as to not be blocked.

The fun bit of this is that this should require zero changes in the client.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,6 @@ is installed by `pip`.
 - Click
 - diskcache
 - inflection
-- isodate
 - Jinja2
 - jsonschema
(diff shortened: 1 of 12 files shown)


mdboom:
I think the proposal is finalized now, so we can take a final look at this.

Dexterp37:
@mdboom once we use this in glean, we should have another PR to make the metrics never expiring (in addition to the parsers bump). This shouldn't involve a new data review, but we should still let Chenxia know I think!

PR #42, opened 2019-03-15
1535339, 1535341: Improvements to event extra metadata

mdboom:
This ties up a couple of loose ends from the event API changes proposal.

1) It makes sure that event extra data keys starting with `glean.` are reserved for internal use.

2) It makes the entries in the extra_keys parameter an object rather than a string so we can attach other metadata in the future.

(It was easiest to handle these two bugs as a unit, since they are so closely related).

### DIFF
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -35,16 +35,26 @@ class Metric:
         super().__init_subclass__(**kwargs)
 
     @classmethod
-    def make_metric(cls, category, name, metric_info, validated=False):
+    def make_metric(cls, category, name, metric_info, config={}, validated=False):
         """
         Given a metric_info dictionary from metrics.yaml, return a metric
         instance.
+
+        :param: category The category the metric lives in
+        :param: name The name of the metric
+        :param: metric_info A dictionary of the remaining metric parameters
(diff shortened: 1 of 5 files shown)


glean_parser/metrics.py
```diff
@@ -240,7 +256,19 @@ class Event(Metric):
 
     @property
     def allowed_extra_keys(self):
-        return list(self.extra_keys.keys())
+        # Sort keys so that output is deterministic
+        return sorted(list(self.extra_keys.keys()))
+
+    @staticmethod
+    def validate_extra_keys(extra_keys, config):
```
Dexterp37: @mdboom not strictly related to this code location, but should we at least standardize over the `description` metadata key, to have parity with the current behaviour? I mean, we should at the very least require a `description` key in the metadata for a specific `extra_key`. What do you think?

mdboom: It is -- it's enforced by the jsonschema in this PR.  The only reason this validation happens here (in Python) is because it's dependent on a flag, which is something JSON schema can't do.

Dexterp37: Mh, interesting. For some weird UI artifact this PR was only showing up changes to 2 files to me. I was missing changes to the YAML file.

PR #44, opened 2019-03-19
1536083: Use enums for extra keys on events

mdboom:
This is the glean_parser side of using `enums` for events.

### DIFF
--- a/glean_parser/templates/kotlin.jinja2
+++ b/glean_parser/templates/kotlin.jinja2
@@ -7,34 +7,45 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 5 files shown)


glean_parser/templates/kotlin.jinja2
```diff
@@ -7,34 +7,45 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+{% macro metric_declaration(metric, suffix='', access='') %}
```
Dexterp37: nit: do we need a blank line above this?

mdboom: I wish we could.  If we have one, you end up with two blank lines in a row in the output and `ktlint` complains :(

Dexterp37: Gah :S Nevermind then!

PR #50, opened 2019-03-29
Update template import statements to add .metrics sub-package

travis79:
This must coincide with changes to the glean library, moving all metric types to a /metrics subdirectory.  I'll update this with the PR # from android-components as soon as it's filed.


travis79:
Once landed, need to land the client [PR](https://github.com/mozilla-mobile/android-components/pull/2566)

PR #58, opened 2019-04-18
1529226: Add support for pings.yaml

mdboom:
This adds support for `pings.yaml` files to `glean_parser`, in support of describing pings so that they can indicate whether they should include the `client_info` block.

### DIFF
--- a/glean_parser/schemas/pings.1-0-0.schema.yaml
+++ b/glean_parser/schemas/pings.1-0-0.schema.yaml
@@ -1,0 +1,53 @@
+$schema: http://json-schema.org/draft-07/schema#
+title: Pings
+description: |
+  Schema for the pings.yaml files for Mozilla's glean telemetry SDK.
+
+  The top-level of the `pings.yaml` file has a key defining the name of each
+  ping. The values contain metadata about that ping. Ping names must be
+  snake_case, though they may also have dots `.` to define subcategories.
+
+$id: moz://mozilla.org/schemas/glean/pings/1-0-0
+
+definitions:
+  dotted_snake_case:
+    type: string
+    pattern: "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})*$"
+    maxLength: 40
+
+type: object
+
+propertyNames:
+  anyOf:
+    - $ref: "#/definitions/dotted_snake_case"
+    - enum: ['$schema']
+
+properties:
+  $schema:
+    type: string
+    format: url
+
+additionalProperties:
+  type: object
+  properties:
+    description:
+      title: Description
+      description: |
+        **Required.**
+
+        A textual description of the purpose of this ping and what it contains.
+      type: string
+
+    include_client_id:
+      title: Include client id
+      description: |
+        **Required.**
+
+        When `true`, include the `client_id` value in the ping.
+      type: boolean
+
+  required:
+    - description
+    - include_client_id
+
+  additionalProperties: false
--- a/glean_parser/templates/kotlin.jinja2
+++ b/glean_parser/templates/kotlin.jinja2
@@ -7,15 +7,15 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-{% macro metric_declaration(metric, suffix='', access='') %}
-{{ access }}val {{ metric.name|camelize }}{{ suffix }}: {{ metric|metric_type_name }} by lazy {
-        {{ metric|metric_type_name }}(
-            {% for arg_name in extra_args if metric[arg_name] is defined %}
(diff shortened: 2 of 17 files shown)


glean_parser/schemas/pings.1-0-0.schema.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+        When `true`, include the `client_info` block in the ping.
+      type: boolean
+
+    disabled:
+      title: Disabled
+      description: |
+        If `true`, the ping is disabled, and will not be sent.
+      type: boolean
+      default: false
```
mdboom: As Travis mentioned in the bug, maybe we could/should add data review-related metadata here as well.  `data_review`, `notification_emails`, `bugs`...?

Dexterp37: Yes, but this is probably material for a proposal document. I suggest we do the following: keep the properties for just the `client_id` and the description, file a follow-up bug for defining which metadata should go in here, have a proposal and let it fly by the rest of the glean team.

glean_parser/schemas/pings.1-0-0.schema.yaml
```diff
@@ -0,0 +1,60 @@
...
+  properties:
+    description:
+      title: Description
+      description: |
+        **Required.**
+
+        A textual description of the purpose of this ping and what it contains.
+      type: string
+
+    include_client_info:
```
Dexterp37: As mentioned in the related a-c PR, we don't need to exclude the whole `client_info`. We only want the ability to toggle the `client_id` on or off, as that section contains other data (app version, app name, etc.) which we'd still need.

glean_parser/schemas/pings.1-0-0.schema.yaml
travis79: +1 for this, and the proposal

tests/data/pings.yaml
```diff
@@ -0,0 +1,9 @@
+customPing:
+    description:
+        This is a custom ping
+    include_client_id: false
+
+reallyCustomPing:
```
Dexterp37: Note that `reallyCustomPing` can't be used in `metrics.yaml` (for `send_in_pings`), because we require [snake case](https://github.com/mozilla/glean_parser/blob/30642e1d91ce7f88cca4beba76dd9f08f6f51383/glean_parser/schemas/metrics.1-0-0.schema.yaml#L175) for entries there. I think we'd need to be consistent between the two (i.e. enforce the same constraints).

glean_parser/templates/kotlin.jinja2
```diff
@@ -7,61 +7,61 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-{% macro metric_declaration(metric, suffix='', access='') %}
-{{ access }}val {{ metric.name|camelize }}{{ suffix }}: {{ metric|metric_type_name }} by lazy {
-        {{ metric|metric_type_name }}(
-            {% for arg_name in extra_args if metric[arg_name] is defined %}
-            {{ arg_name|camelize }} = {{ metric[arg_name]|kotlin }},
+{% macro obj_declaration(obj, suffix='', access='', lazy=False) %}
```
Dexterp37: nit: would it make sense to have 2 different templates? we're importing some things here that are not really useful for pings. What do you think?

glean_parser/schemas/pings.1-0-0.schema.yaml
```diff
@@ -0,0 +1,53 @@
+$schema: http://json-schema.org/draft-07/schema#
+title: Pings
+description: |
+  Schema for the pings.yaml files for Mozilla's glean telemetry SDK.
+
+  The top-level of the `pings.yaml` file has a key defining the name of each
+  ping. The values contain metadata about that ping. Ping names must be
```
Dexterp37: This seems to contrast with the `tests/data/pings.yaml`, which is currently using *camelCase*. Is the schema validation broken?

tests/data/pings.yaml
mdboom: Good catch.  Definitely the intention is to use snake case here.  The unit test for catching this was wrong...  Should be fixed now.

glean_parser/templates/kotlin.jinja2
mdboom: Given the complexity of the template just for the basics of generating Kotlin, I'd rather keep it to one template for everything.  Plus, pings aren't really a special case -- they are just another metric type as far as `glean_parser` is concerned.

glean_parser/schemas/pings.1-0-0.schema.yaml
mdboom: The test was broken -- but validation is actually working.  Should be fixed (and enforced correctly) now.

glean_parser/templates/kotlin.jinja2
Dexterp37: Got it, ok!

PR #61, opened 2019-04-30
1547967: Add additional metadata to pings

mdboom:
Adds the metadata related to data review and expiration from metrics to pings as well.

### DIFF
--- a/glean_parser/schemas/pings.1-0-0.schema.yaml
+++ b/glean_parser/schemas/pings.1-0-0.schema.yaml
@@ -46,8 +46,56 @@ additionalProperties:
         When `true`, include the `client_id` value in the ping.
       type: boolean
 
+    notification_emails:
+      title: Notification emails
+      description: |
+        **Required.**
+
+        A list of email addresses to notify for important events with the
+        ping or when people with context or ownership for the ping need to
+        be contacted.
+      type: array
+      minItems: 1
+      items:
+        type: string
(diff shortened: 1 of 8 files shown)


glean_parser/schemas/pings.1-0-0.schema.yaml
```diff
@@ -46,8 +46,79 @@ additionalProperties:
...
+        format: uri
+
+    disabled:
+      title: Disabled
+      description: |
+        If `true`, the ping is disabled, and attempting to send it is a no-op.
+      type: boolean
+      default: false
+
+    expires:
```
Dexterp37: I think @georgf had strong opinions on this one. Let's make sure to check with him before merging this.

georgf: Yeah, i'd love to have a design conversations on this specific feature first.
I'm not sure how this should work together with probe expiry attributes without surprising users.
I also don't understand yet what a typical user story around a time limited ping is.

For me my model is currently more that pings are more transport - as such it matters what metrics in them expire or not, not that the ping expires.

Unless we're already clear on the above points (maybe i just missed that?), i would like us to avoid adding the complexity here and think about the impact first.

mdboom:
I've ripped out the expiration / disabling on pings, which can definitely be addressed later, in order to move this forward.

PR #69, opened 2019-05-22
Allow dotted snake-case for labeled metric IDs

travis79:
This expands the regex for labeled metrics to allow for dotted snake-case.


travis79:
Should wait for pipeline-schema [PR](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/335) to be deployed before we deploy this

mdboom:
I think now that we have notification of this being deployed, this is good-to-merge?

PR #70, opened 2019-05-30
1553186: Make pings metadata fields required.

mdboom:
Now that we have these fields everywhere that has a pings.yaml, let's make them officially required.  Any future custom pings will be forced to set these which is good for data stewardship.


mdboom:
Testing is failing on installing a Debian package.  Hopefully that's just intermittent.

PR #73, opened 2019-07-17
Allow for '-' in metrics labels

Dexterp37:
This is related to mozilla-services/mozilla-pipeline-schemas#356 . It's the second part to the fix for [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1566764).


mdboom:
We should add some example labels to `tests/data/core.yaml`.  There aren't any now (there should have been) but maybe let's add them while we're here.

PR #75, opened 2019-07-22
 Enable parsing of the `gecko_datapoint` attribute in metrics.yaml 

Dexterp37:
This adds support for the `gecko_datapoint` property in the `metrics.yaml` file for the Glean SDK.

In addition, this updates the jinja2 template to produce a function to map Gecko metrics to Glean SDK metrics and adds test coverage.

See [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1566356) and the blocked meta bug for more context.

### DIFF
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -208,6 +208,9 @@ class Timespan(TimeBase):
 @dataclass
 class TimingDistribution(TimeBase):
     typename = 'timing_distribution'
+    # The following is a Gecko-specific property for using the Glean SDK
+    # with GeckoView metrics. See bug 1566356 for more context.
+    gecko_datapoint: str = ''
 
 
 @dataclass
--- a/glean_parser/schemas/metrics.1-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.1-0-0.schema.yaml
@@ -333,6 +333,17 @@ definitions:
             - description
         default: {}
 
+      gecko_datapoint:
+        title: Gecko Datapoint
+        description: |
+          This is a Gecko-specific property. It is the name of the Gecko histogram to
+          accumulate the data from, when using the Glean SDK in a product using GeckoView.
+          See bug 1566356 for more context.
+
+          This is valid when `type`_ is `timing_distribution`.
+
+        type: string
+
     required:
       - type
       - bugs
--- a/glean_parser/templates/kotlin.jinja2
+++ b/glean_parser/templates/kotlin.jinja2
@@ -23,6 +23,9 @@ package {{ namespace }}
 import {{ glean_namespace }}.private.Lifetime // ktlint-disable no-unused-imports
 import {{ glean_namespace }}.private.TimeUnit // ktlint-disable no-unused-imports
 import {{ glean_namespace }}.private.NoExtraKeys // ktlint-disable no-unused-imports
+{% if has_gecko_datapoints %}
+import {{ glean_namespace }}.private.HistogramBase
+{% endif %}
 {% for obj_type in obj_types %}
 import {{ glean_namespace }}.private.{{ obj_type }}
 {% endfor %}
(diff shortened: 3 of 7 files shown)


glean_parser/metrics.py
```diff
@@ -208,6 +208,9 @@ class Timespan(TimeBase):
 @dataclass
 class TimingDistribution(TimeBase):
     typename = 'timing_distribution'
+    # The following is a Gecko-specific property for using the Glean SDK
+    # with GeckoView metrics. See bug 1566356 for more context.
+    gecko_datapoint: str = ''
```
mdboom: Do we want this only on `TimingDistribution`?  Theoretically at least, it might appear on other metric types.

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -333,6 +333,17 @@ definitions:
...
         default: {}
 
+      gecko_datapoint:
+        title: Gecko Datapoint
+        description: |
+          This is a Gecko-specific property. It is the name of the Gecko histogram to
+          accumulate the data from, when using the Glean SDK in a product using GeckoView.
+          See bug 1566356 for more context.
+
+          This is valid when `type`_ is `timing_distribution`.
```
mdboom: This new parameter also needs an entry in `docs/metrics-yaml.rst` so it will show up in the docs.

glean_parser/templates/kotlin.jinja2
```diff
@@ -64,4 +67,18 @@ internal object {{ category_name|Camelize }} {
...
+        return when(geckoMetricName) {
+        {% for obj in objs.values() %}
+            {% if obj.gecko_datapoint %}
+            "{{ obj.gecko_datapoint }}" -> {{ obj.name|camelize }}
+            {% endif %}
+        {%- endfor %}
+            else -> null
+        }
+    }
+{% endif %}
```
mdboom: This is fine, but wanted to call out that since each category of metrics is written to its own `.kt` file, this function will only include the mapping for a particular category.  Maybe that's sufficient for the use case, of course.

If we needed a global mapping of everything, we'd probably need to use a technique similar to how pings are registered -- i.e. a singleton data structure that is updated at startup time.

tests/test_kotlin.py
```diff
@@ -174,3 +174,55 @@ def test_glean_namespace(tmpdir):
...
+        # Make sure we're adding the relevant Glean SDK import, once.
+        assert content.count(
+            "import Bar.private.HistogramBase"
+        ) == 1
+
+        # Validate the generated Gecko metric mapper Kotlin operator.
+        # NOTE: Indentation, whitespaces  and text formatting of the block
+        # below are important. Do not change them unless the file format
+        # changes, otherwise validation will fail.
+        expected_operator = """    operator fun get(geckoMetricName: String): HistogramBase? {
```
mdboom: If we wanted to make this more robust to whitespace changes, you could use a regex to remove all whitespace from both the reference and test strings (I think there are other tests that do that).  But not a requirement for this PR, IMHO.

tests/test_kotlin.py
```diff
@@ -174,3 +174,55 @@ def test_glean_namespace(tmpdir):
...
+        assert expected_operator in content
+
+    with open(tmpdir / 'NonGeckoMetrics.kt', 'r', encoding='utf-8') as fd:
+        content = fd.read()
+        assert 'HistogramBase' not in content
+
+    # Only run this test if ktlint is on the path
+    if shutil.which('ktlint'):
+        for filepath in tmpdir.glob('*.kt'):
+            subprocess.check_call(['ktlint', filepath])
```
mdboom: :+1:  Our tests are thorough... :)

glean_parser/metrics.py
Dexterp37: This will be used by the `memory_distribution` as well, once we'll have that. We might want to reconsider where this lives once that lands

tests/test_kotlin.py
Dexterp37: I just realized that this *already* saved my back :D

tests/test_kotlin.py
Dexterp37: Thanks, I tried to do this and it failed due to the nested indentation of the block :o Anyway, given that it's not a requirement, I'll happily skip over this :P

glean_parser/templates/kotlin.jinja2
Dexterp37: As we discussed over Slack, let's take this as follow-up work since it's not blocking the current batch of metrics. I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1567894).

PR #79, opened 2019-07-25
Enable the use of multiple categories for Gecko metrics

Dexterp37:
This introduces a new template used for generating a single mapping object. The object then calls in the
other generated metrics, if needed.


glean_parser/kotlin.py
```diff
@@ -163,9 +227,11 @@ def output_kotlin(objs, output_dir, options={}):
...
                     namespace=namespace,
                     has_labeled_metrics=has_labeled_metrics,
-                    has_gecko_datapoints=has_gecko_datapoints,
                     glean_namespace=glean_namespace,
                 )
             )
             # Jinja2 squashes the final newline, so we explicitly add it
             fd.write('\n')
+
+    # TODO: Maybe this should just be a separate outputter?
```
Dexterp37: Note: creating a custom outputter would require duplicating some of the kotlin code anyway. It would also need changes on a-c to call in the correct outputter when in engine-gecko-*

glean_parser/templates/kotlin.geckoview.jinja2
```diff
@@ -0,0 +1,34 @@
...
+ */
+internal object GleanGeckoHistogramMapping {
+    // Support exfiltration of Gecko metrics from products using both the
+    // Glean SDK and GeckoView. See bug 1566356 for more context.
+    operator fun get(geckoMetricName: String): HistogramBase? {
+        return when (geckoMetricName) {
+        {% for category in gecko_metrics.keys()|sort %}
+            // From {{ category|Camelize }}.kt
+            {% for metric in gecko_metrics[category] %}
+            "{{ metric.gecko_datapoint }}" -> {{ category|Camelize }}.{{ metric.name|camelize }}
```
mdboom: I'm not sure it matters, but I think this means all of the (Gecko) histograms will be instantiated when you use the first one.

glean_parser/templates/kotlin.geckoview.jinja2
```diff
@@ -0,0 +1,34 @@
...
+@file:Suppress("PackageNaming")
+package {{ namespace }}
+
+import {{ glean_namespace }}.private.HistogramBase // ktlint-disable import-ordering no-unused-imports
+
+/*
+ * This class performs the mapping between Gecko Histograms and Glean SDK
+ * metric types.
+ */
+internal object GleanGeckoHistogramMapping {
```
mdboom: So, I think this implementation still means that there is one mapping per library -- not Glean (app) wide.  I think that's fine here, just wanted to call that out.  It also means (I think) that only engine-gecko has access to this mapping, not Glean itself (unless it's passed into Glean somehow)...

glean_parser/templates/kotlin.geckoview.jinja2
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@file:Suppress("PackageNaming")
+package {{ namespace }}
+
+import {{ glean_namespace }}.private.HistogramBase // ktlint-disable import-ordering no-unused-imports
```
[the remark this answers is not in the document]
Dexterp37: No, because they are from the same package. I assumed so too, but then `ktlint` started complaining about useless imports.

glean_parser/templates/kotlin.geckoview.jinja2
Dexterp37: Mh, I tried this (check [here](https://pl.kotl.in/cnrMCzHIV) for a PoC) and doesn't look like it's triggering the load of other histograms. Is that what you meant?

glean_parser/templates/kotlin.geckoview.jinja2
Dexterp37: Yes, you are correct. Given the specifics of this, this is fine: there should be *one* GeckoView component embedding Gecko, not multiple per app. For context, we will create a new `Gecko.RuntimeTelemetry.Delegate` implementation in `engine-gecko-nightly`: this delegate will be the only one instance interacting with `GleanGeckoHistogramMapping`, as shown below:

```
// TODO: Change RemoveMe.Delegate to RuntimeTelemetry.Delegate once the
// GV API from Bug 1566367 is merged.
class GeckoAdapter : RemoveMe.Delegate {
    override fun onTelemetryReceived(metric: RemoveMe.Metric) {
        GleanGeckoHistogramMapping[metric.name]?.accumulateSamples(metric.values)
    }

}
```

glean_parser/templates/kotlin.geckoview.jinja2
mdboom: Yep -- that's what I meant.  Thanks for confirming

PR #80, opened 2019-07-29
Bug 1564486 - Disallow labeled timespans

badboy:


badboy:
Unfortunately removing labeled timespans just in the parser means I have to touch the schema here. Is that still okay without a version bump? Or can we version bump?

PR #82, opened 2019-07-30
Add instructions to use changelog, and backfill changelog

mdboom:
Yesterday, I tagged a 1.0.0 release, and with that comes strictly using semver and keeping a good changelog.

This just adds that to the PR template, and backfills the changelog with one thing that slipped through the cracks yesterday.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,8 @@
 History
 =======
 
+* `labeled_timespan` is no longer an allowed metric type.
+
 1.0.0 (2019-07-29)
 ------------------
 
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,8 @@
 History
 =======
```
badboy: Add a header here?

```
Unreleased
----------
```

mdboom:
> Where to increase the version number

Don't need to.  It's pulled from git, thanks to `setuptools-scm` package.  Added a note about that anyway, because it's a bit surprising.

PR #83, opened 2019-08-06
1571744: Add custom_distribution metric type

mdboom:

### DIFF
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -11,7 +11,7 @@ Classes for each of the high-level metric types.
 import dataclasses
 from dataclasses import dataclass, field, InitVar
 import enum
-from typing import Dict, List, Set, Tuple, Union
+from typing import Dict, List, Set, Union
 
 from . import parser
 from . import util
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -225,10 +225,10 @@ class HistogramType(enum.Enum):
 class CustomDistribution(Metric):
     typename = 'custom_distribution'
 
+    range_min: int = 1
     # The defaults must be provided here to work with Python's @dataclass,
     # however, in practice, these parameters are required by the schema,
     # so they will never be used.
-    range_min: int = -1
     range_max: int = -1
     bucket_count: int = -1
     histogram_type: HistogramType = HistogramType.exponential
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -11,7 +11,7 @@ Classes for each of the high-level metric types.
 import dataclasses
 from dataclasses import dataclass, field, InitVar
 import enum
-from typing import Dict, List, Set, Union
+from typing import Dict, List, Set, Tuple, Union
 
 from . import parser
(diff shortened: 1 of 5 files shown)


glean_parser/metrics.py
```diff
@@ -208,9 +214,20 @@ class Timespan(TimeBase):
...
+class HistogramType(enum.Enum):
+    linear = 0
+    exponential = 1
+
+
+@dataclass
+class CustomDistribution(Metric):
+    typename = 'custom_distribution'
+
+    range: Tuple[int, int] = field(default_factory=lambda: [0, 60000])
```
Dexterp37: Any specific reason why we're picking these defaults here, in the parser?

glean_parser/metrics.py
```diff
@@ -208,9 +214,20 @@ class Timespan(TimeBase):
...
+
+
+class HistogramType(enum.Enum):
+    linear = 0
+    exponential = 1
+
+
+@dataclass
+class CustomDistribution(Metric):
+    typename = 'custom_distribution'
```
Dexterp37: This might require an update depending if you want to change the constructor in mozilla-mobile/android-components#4041

glean_parser/metrics.py
mdboom: Oh -- since these were the defaults picked for the original timing distribution implementation, I thought they were somehow the defaults on desktop.  But it looks from [here](https://searchfox.org/mozilla-central/source/toolkit/components/telemetry/build_scripts/mozparsers/parse_histograms.py) like the only default is that `rangeMin` is 1.

mdboom:
Desktop telemetry limits `bucket_count` to 100, unless the metric is in an allow list.  Should we do that here?  (Easy enough to set a maximum -- don't know if I want to take on the complexity of supporting an allow list, though).

glean_parser/metrics.py
Dexterp37: Should we have the defaults on the types in a-c ? Also, yes, let's keep the same defaults. Documentation for the desktop part [lives here](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/histograms.html#low)

Dexterp37:
> 
> 
> Desktop telemetry limits `bucket_count` to 100, unless the metric is in an allow list. Should we do that here? (Easy enough to set a maximum -- don't know if I want to take on the complexity of supporting an allow list, though).

Yes please, let's limit to 100. We don't need an allow list, we only have one due to legacy telemetry on Desktop.

glean_parser/metrics.py
mdboom: Yeah -- I've updated this to have *no defaults*.  I guess to be fully consistent with desktop, though, we would allow `rangeMin` to have a default of 1.

We seem to have a mix of providing defaults in a-c and not.  I think it's best to not have them in a-c, since they wouldn't matter in practice anyway -- glean_parser always generates a value for every parameter when it outputs Kotlin code.  The Kotlin-side defaults would only affect writing Glean's own unit tests.

PR #84, opened 2019-08-12
Remove labeled string list & UUID

badboy:
Draft while I double-check the metric file once more.

### DIFF
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -288,11 +288,6 @@ class LabeledString(Labeled, String):
 
 
 @dataclass
-class LabeledStringList(Labeled, StringList):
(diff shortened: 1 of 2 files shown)


badboy:
I looked at:
* android-components
* application-services
* fenix
* reference-browser

None contain labeled metrics other than `labeled_counter` and `labeled_string`

mdboom:
In fairness, [not untested](https://github.com/mozilla-mobile/android-components/blob/284b533ffe5537fa03ef78db0c8737772a4d3f85/components/service/glean/src/test/java/mozilla/components/service/glean/private/LabeledMetricTypeTest.kt#L371), just unused.

PR #85, opened 2019-08-12
Memory histogram

mdboom:

### DIFF
--- a/glean_parser/schemas/metrics.1-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.1-0-0.schema.yaml
@@ -103,6 +103,9 @@ definitions:
 
             - `uuid`: Record a UUID v4.
 
+            - `memory_distribution`: A histogram for recording memory usage values.
+              Additional properties: `memory_unit`_.
+
             - `custom_distribution`: A histogram with a custom range and number of buckets.
               This metric type is for legacy support only and is only allowed for
               metrics coming from GeckoView.
@@ -129,6 +132,7 @@ definitions:
           - timespan
           - timing_distribution
           - custom_distribution
+          - memory_distribution
           - datetime
           - use_counter
           - usage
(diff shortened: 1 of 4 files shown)


glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -294,6 +298,24 @@ definitions:
           - hour
           - day
 
+      memory_unit:
+        title: Memory unit
+        description: |
+          The unit that the incoming memory size values are recorded in.
+
+          The units are the power-of-2 units, so "kilobyte" is pedantically a "kibibyte".
```
chutten: Well actually a more accurate adjective would be "correctly". \*pushes bridge of glasses up nose*

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -294,6 +298,24 @@ definitions:
...
+
+          - kilobyte == 2^10 ==         1,024 bytes
+          - megabyte == 2^20 ==     1,048,576 bytes
+          - gigabyte == 2^30 == 1,073,741,824 bytes
+
```
chutten: Include a note about truncation/rounding? (Even if there is none)

glean_parser/schemas/metrics.1-0-0.schema.yaml
mdboom: Indeed.  That's probably a better (and less condescending) word to use.

glean_parser/schemas/metrics.1-0-0.schema.yaml
mdboom: Good point.

glean_parser/schemas/metrics.1-0-0.schema.yaml
chutten: And here I was saying it as a joke.

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -450,7 +474,11 @@ additionalProperties:
                 enum:
                   - timing_distribution
                   - custom_distribution
+                  - memory_distribution
```
Dexterp37: Should we update `test_geckoview_only_on_valid_metrics` to cover this too?

PR #90, opened 2019-08-28
Restrict ping names to be kebab-case

jklukas:
See background discussion in https://github.com/mozilla-services/mozilla-pipeline-schemas/issues/390

The only known cases of existing glean pings that do not already match this naming convention are in the [android components sync library metrics.yaml](https://github.com/mozilla-mobile/android-components/blob/master/components/browser/storage-sync/metrics.yaml) so my main concern is understanding exactly what the upgrade path is.

If this change is merged in the glean parser, will that start failing builds for android components and/or apps using the sync library? Do we need to stage changes for that whole stack and merge concurrently with this change?

Alternately, should we bump the schema version for pings.schema.yamlto ease the transition?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,8 @@ History
 Unreleased
 ----------
 
+* Restrict ping names to be kebab-case and change `all_pings` to `all-pings`
+
 1.4.1 (2019-08-28)
 ------------------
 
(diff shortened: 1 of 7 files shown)


jklukas:
Note that the updates in MSG and MPS have not yet been made. Let me know if those need to be prioritized to support having this merged to master.

PR #101, opened 2019-09-16
Create metrics for Swift

badboy:
This is not yet fully tested.
Events and labeled metrics are not yet implemented, so their exact API is not yet defined.


tests/test_swift.py
```diff
@@ -0,0 +1,116 @@
...
+    assert kdf([42, "\n"]) == r'[42, "\n"]'
+    assert (
+        kdf({"key": "value", "key2": "value2"})
+        == r'["key": "value", "key2": "value2"]'
+    )
+    assert kdf(metrics.Lifetime.ping) == ".ping"
+
+
+def test_metric_type_name():
+    event = metrics.Event(
```
Dexterp37: As far as I can see, the only metric type available in the iOS implementation is the `CounterMetricType`. Let's just account for it, leaving all the other types out: we can land this as an initial PR and iterate on it as needed.

tests/test_swift.py
```diff
@@ -0,0 +1,116 @@
...
+        name="metric",
+        bugs=[42],
+        notification_emails=["nobody@example.com"],
+        description="description...",
+        expires="never",
+    )
+
+    assert swift.type_name(event) == "EventMetricType<noExtraKeys>"
+
+    boolean = metrics.Boolean(
```
Dexterp37: Let's leave booleans out since iOS doesn't seem to support them for now.

glean_parser/templates/swift.jinja2
```diff
@@ -0,0 +1,63 @@
...
+        case {{ key|camelize }}
+    {% endfor %}
+    }
+    {% endif %}
+{% endfor %}
+{% for obj in objs.values() %}
+    {% if obj.labeled %}
+    {{ obj_declaration(obj, 'Label', 'private ') }}
+    /// {{ obj.description|wordwrap(wrapstring='\n     /// ') }}
+    static let {{ obj.name|camelize }}: LabeledMetricType<{{ obj|type_name }}> = LabeledMetricType(
```
Dexterp37: We don't know yet how labeled things will work in the iOS API, don't we? I'd be for leaving things that do not have a clear implementation already available out of this PR.

badboy:
I explicitly removed support for labeled metrics and events.
Other metrics are simpler, so there's no harm in just keeping them for now.

badboy:
`CorePing.swift` (from `core.yaml`):

```swift
// -*- mode: Swift -*-

import Glean

internal enum CorePing {
    /// Running ping counter for this ping. This works as a sequence number to allow
    /// detecting for missing send_in_pings.
    static let seq = CounterMetricType(
        category: "core_ping",
        name: "seq",
        sendInPings: ["core"],
        lifetime: .application,
        disabled: false
    )

    /// The date the ping was created.
    static let created = DatetimeMetricType(
        category: "core_ping",
        name: "created",
        sendInPings: ["core"],
        lifetime: .ping,
        disabled: false,
        timeUnit: .day
    )

    /// The number of sessions since the last upload.
    static let sessions = CounterMetricType(
        category: "core_ping",
        name: "sessions",
        sendInPings: ["core"],
        lifetime: .ping,
        disabled: false
    )

    /// The combined duration, in seconds, of all the sessions since the last upload.
    static let durations = TimespanMetricType(
        category: "core_ping",
        name: "durations",
        sendInPings: ["core"],
        lifetime: .ping,
        disabled: false,
        timeUnit: .second
    )

    /// The number of times the flash plugin was used since the last upload.
    static let flashUsage = CounterMetricType(
        category: "core_ping",
        name: "flash_usage",
        sendInPings: ["core"],
        lifetime: .ping,
        disabled: false
    )

    /// True if the user has set Firefox as the default browser.
    static let defaultBrowser = BooleanMetricType(
        category: "core_ping",
        name: "default_browser",
        sendInPings: ["core"],
        lifetime: .ping,
        disabled: false
    )

}
```

`Ping.swift` (from `pings.yaml`):

```swift
// -*- mode: Swift -*-

import Glean

internal class Pings {
  public let shared = Pings()
  private init() {
    // intentionally left private, no external user can instantiate a new global object.
  }

  /// This is a custom ping
  let customPing = Ping(name: "custom_ping", includeClientId: false)
  /// This is another custom ping
  let reallyCustomPing = Ping(name: "really_custom_ping", includeClientId: true)
}
```

glean_parser/templates/swift.jinja2
```diff
@@ -0,0 +1,67 @@
...
+
+// AUTOGENERATED BY glean_parser.  DO NOT EDIT.
+{# The rendered markdown is autogenerated, but this
+Jinja2 template is not. Please file bugs! #}
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+{% macro obj_declaration(obj, suffix='', access='') %}
+{{ access }}static let {{ obj.name|camelize }}{{ suffix }} = {{ obj|type_name }}(
```
Dexterp37: I know that I already asked for this, but I can't really find the answer. Shouldn't this be lazy?

glean_parser/templates/swift.jinja2
```diff
@@ -0,0 +1,67 @@
...
+    )
+{% endmacro %}
+
+import Glean
+
+{% if is_ping_type %}
+internal class {{ category_name|Camelize }} {
+  public let shared = {{ category_name|Camelize }}()
+  private init() {
+    // intentionally left private, no external user can instantiate a new global object.
```
Dexterp37: nit: leading capital letter "I"?

tests/test_swift.py
```diff
@@ -0,0 +1,112 @@
...
+    with open(tmpdir / "Telemetry.swift", "r", encoding="utf-8") as fd:
+        content = fd.read()
+        assert "جمع 搜集" in content
+
+    with open(tmpdir / "GleanInternalMetrics.swift", "r", encoding="utf-8") as fd:
+        content = fd.read()
+        assert 'category: ""' in content
+
+    # Only run this test if swiftlint is on the path
+    if shutil.which("swiftlint"):
```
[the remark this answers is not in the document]
badboy: Good point, happy to try doing that in a followup (not sure how easy it is to get swiftlint on Linux, I'd rather not have another build target)

glean_parser/templates/swift.jinja2
badboy: > Global constants and variables are always computed lazily, in a similar manner to Lazy Stored Properties. Unlike lazy stored properties, global constants and variables do not need to be marked with the lazy modifier.

That's what https://docs.swift.org/swift-book/LanguageGuide/Properties.html says.

Dexterp37: This makes the container `lazy` by default (e.g. `CorePing`), but does it make members of it lazy too (e.g. `seq`)?

tests/test_swift.py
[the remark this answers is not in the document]
Dexterp37: Sure! Let's file a bug about this :)

glean_parser/templates/swift.jinja2
badboy: `CorePing` is an empty enum, it can't be instantiated (it merely acts as a namespace.

I'm pretty sure the other things are therefore "global constants" and lazily evaluated. Let me quickly verify that.

Dexterp37: Thanks. Sorry for being pedantic on this, but it is quite important for all the metrics to be lazy. We don't want `CorePing.seq` to be instantiated if we access `CorePing.defaultBrowser`.

badboy: Run this test on a minimal version of that `CorePing.swift` and a bit more logging for metric instantiation:

```swift
    func testMetricLazyness() {
        print("starting lazyness test")
        let glean = Glean.shared
        glean.initialize()
        print("Glean initialized. Incremeting sequence number")
        CorePing.seq.add()
        print("CorePing.seq incremented. Next: flash usage")
        CorePing.flashUsage.add(2)
    }
```

Got this output (stripped of unnecessary information:

```
[INFO  glean_core] Creating new Glean
[INFO  glean_core::database] Database initialized
[INFO  glean_ffi] Glean initialized
Glean initialized. Incremeting sequence number
[INFO  glean_ffi::counter] Creating new metric of type 'CounterMetric' with name: seq
CorePing.seq incremented. Next: flash usage
[INFO  glean_ffi::counter] Creating new metric of type 'CounterMetric' with name: flash_usage
```

As you can see they are lazy and get created on first access.

Dexterp37: Great, thanks!

tests/test_swift.py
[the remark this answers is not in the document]
mdboom: Yeah -- doesn't look easy:  https://github.com/realm/SwiftLint/issues/732

glean_parser/swift.py
```diff
@@ -0,0 +1,155 @@
...
+        "allowed_extra_keys",
+        "bucket_count",
+        "denominator",
+        "histogram_type",
+        "include_client_id",
+        "range_max",
+        "range_min",
+        "time_unit",
+        "values",
+    ]
```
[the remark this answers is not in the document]
badboy: I trimmed down the list to only what's currently needed. We can re-evaluate moving it out once it becomes the same as for Kotlin.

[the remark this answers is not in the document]
mdboom: WFM.  It would be nice to unify down the road, since this getting out-of-sync has been the source of bugs in the past...

badboy:
Filed https://github.com/mozilla/glean_parser/issues/103

PR #115, opened 2019-10-04
1585953: Warn when using bug numbers

mdboom:

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,8 @@ History
 Unreleased
 ----------
 
+* Glinter now warns when bug numbers (rather than URLs) are used.
+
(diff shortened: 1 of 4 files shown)


HISTORY.rst
```diff
@@ -5,6 +5,8 @@ History
 Unreleased
 ----------
 
+* Glinter now warns when bug numbers (rather than URLs) are used.
```
Dexterp37: Do we need to update at least the description [here](https://github.com/mozilla/glean_parser/blob/9c78d030b0a4515384609d12e029c55e47062b2c/glean_parser/schemas/metrics.1-0-0.schema.yaml#L208-L216)? I guess we can't simply change the type there, as we need to allow for people to fix this before enforcing it..

mdboom: Good point.  I've updated the doc to explain that it's deprecated.

PR #123, opened 2019-10-10
New lint: Restrict what metric can go into the 'baseline' ping

badboy:
Essentially only Glean-internal metrics are allowed to go in, that is:
only when the glean_parser is lauched with `--allow-reserved`.


glean_parser/lint.py
```diff
@@ -130,16 +130,30 @@ def check_bug_number(metric):
         )
 
 
+def check_valid_in_baseline(metric, parser_config={}):
+    allow_reserved = parser_config.get("allow_reserved", False)
+
+    if not allow_reserved and "baseline" in metric.send_in_pings:
+        yield (
+            f"The baseline ping is Glean-internal. "
+            "User metrics should go into the 'metrics' ping or custom pings."
```
chutten: Reminds me of all the effort we put into adding links to the docs on parser errors in mozilla-central. Maybe we should do that here (and maybe that should be its own thing).

I mention this because the way to use the 'metrics' ping is to specify nothing at all. And because someone could be adding an event to the baseline ping, and those should go into the 'events' ping. Pointing to the docs where that level of nuance could be explored at length would help for this.

mdboom: :+1:

mdboom:
@chutten: Yes -- currently this will print a warning, but there is a bug to make these all errors in the near future.  Since it's all warnings, we can either "grandfather" it in with `no_lint` -- if it's really necessary to maintain data continuity -- or just fix it in Fenix.

chutten:
> @chutten: Yes -- currently this will print a warning, but there is a bug to make these all errors in the near future. Since it's all warnings, we can either "grandfather" it in with `no_lint` -- if it's really necessary to maintain data continuity -- or just fix it in Fenix.

As you might expect I'm for the latter option: get those out of the "baseline" ping.

glean_parser/lint.py
chutten: While I was reading this again, do we have a lint to prevent people from specifying just "metrics" for a metric? Do we care?

Do we have a lint for specifying "metrics" for an event? Or "events" for a non-event metric?

Do we have a lint for having stores/pings called "metric" or "event" or other one-off misspellings? Even if it's deliberate that the app wants a ping called "metric", we should probably discourage it as it'll end up in a very similarly-named dataset with enough similar information (ping_info and client_info at the least) that there could be confusion.

mdboom: > do we have a lint to prevent people from specifying just "metrics" for a metric? Do we care?

We don't currently.  One could argue that it would be nice to not overspecify so we can change the default behavior easier in the future.  But putting `metrics` in the `send_in_pings` can't be an error, because that's the only way of saying "on the metrics ping and *also* on this other ping" with `[metrics, other]`.

> Do we have a lint for specifying "metrics" for an event? Or "events" for a non-event metric?

Currently, `metrics` and `events` are just the default pings for metrics and events respectively.  I wonder whether or not it should be disallowed -- it's sort of core to the design that metrics of all types can be mixed arbitrarily.  You might want to add a metric that counts the number of events sent to Leanplum on the events ping to make sure its sent on the same cadence as the other events there, for example.

> Do we have a lint for having stores/pings called "metric" or "event" or other one-off misspellings?

That seems like a great idea.  :paperclip: "You said `metric` did you mean `metrics`"?  Let's file a bug for that one :)

Dexterp37:
> 
> 
> > @chutten: Yes -- currently this will print a warning, but there is a bug to make these all errors in the near future. Since it's all warnings, we can either "grandfather" it in with `no_lint` -- if it's really necessary to maintain data continuity -- or just fix it in Fenix.
> 
> As you might expect I'm for the latter option: get those out of the "baseline" ping.

Given that in Fenix is search metrics and that there might be discontinuities in dashboards they are using, I think we should print a warning first (so move on with this) and _encourage_ them to fix this.

badboy:
So ... what's the verdict? Merge, wait for warnings in Fenix and see how they react?

mdboom:
I'm fine with merging this.  We'll probably want to have [warnings and errors](https://bugzilla.mozilla.org/show_bug.cgi?id=1589635) as distinct things someday, rather than (as originally planned) just flip a big "all warnings become errors" switch in the future.

PR #125, opened 2019-10-15
Warn for slight misspellings in ping names.

badboy:
This only handles builtin pings (and doesn't do baseline, because we actually don't want more metrics in there).

### DIFF
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -26,9 +26,7 @@ def _hamming_distance(str1, str2):
 
     diffs = 0
     if len(str1) < len(str2):
-        tmp = str1
-        str1 = str2
-        str2 = tmp
+        str1, str2 = str2, str1
     len_dist = len(str1) - len(str2)
     str2 += " "*len_dist
 
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -18,6 +18,26 @@ def _split_words(name):
     return re.split("[._]", name)
 
 
+def _hamming_distance(str1, str2):
+    """
+    Count the # of differences between strings str1 and str2,
+    padding the shorter one with whitespace
+    """
+
+    diffs = 0
+    if len(str1) < len(str2):
+        tmp = str1
+        str1 = str2
+        str2 = tmp
+    len_dist = len(str1) - len(str2)
+    str2 += " "*len_dist
+
+    for ch1, ch2 in zip(str1, str2):
+        if ch1 != ch2:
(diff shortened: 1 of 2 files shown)


glean_parser/lint.py
```diff
@@ -130,16 +150,41 @@ def check_bug_number(metric):
         )
 
 
+def check_valid_in_baseline(metric, parser_config={}):
```
mdboom: Should we merge #123 first?

glean_parser/lint.py
```diff
@@ -18,6 +18,26 @@ def _split_words(name):
...
+    """
+    Count the # of differences between strings str1 and str2,
+    padding the shorter one with whitespace
+    """
+
+    diffs = 0
+    if len(str1) < len(str2):
+        tmp = str1
+        str1 = str2
+        str2 = tmp
```
mdboom: Could do `str1, str2 = str2, str1`

glean_parser/lint.py
```diff
@@ -130,16 +150,41 @@ def check_bug_number(metric):
...
+def check_misspelled_pings(metric, parser_config={}):
+    builtin_pings = ["metrics", "events"]
+
+    for ping in metric.send_in_pings:
+        for builtin in builtin_pings:
+            distance = _hamming_distance(ping, builtin)
+            if distance == 1:
+                yield (f"Ping '{ping}' seems misspelled. Did you mean '{builtin}'?")
+
+
```
mdboom: I had thought we could do this for all of the "enumeration-like" parameters: `type`, `lifetime`, `time_unit`, `memory_unit`, and `histogram_type`.  Up to you whether you want to do that now or in a follow-on bug.

glean_parser/lint.py
badboy: Oh yes, forgot that I based it on that branch

glean_parser/lint.py
badboy: hm, guess that would make sense. I can take it here.

mdboom: Actually, my suggestion doesn't make much sense.  These others are exactly required already.  :)

chutten: Where it might get interesting is for things where we allow people to write their own strings. Like `custom_distribution`'s "unit" or any `category` name

mdboom: Spell checker?  (j/k)

badboy: Sounds like something for a followup

PR #127, opened 2019-10-18
Make labels from Labeled types sets again

brizental:
These labels were recently changed to lists to ensure the labels
ordering from `metrics.yaml` was kept in the markdown generated
documentation. That had the unwanted effect of changing the type
of the labels in the Kotlin and Swift generated codes.

To fix this, a new property `ordered_labels` was added to the
Labeled types which is a list. Keeping the actual `labels` property
a set.

See discussion: https://github.com/mozilla-mobile/android-components/pull/4750#discussion_r336030971

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -11,6 +11,8 @@ Unreleased
 
 * New lint: Warn for slight misspellings in ping names
 
+* BUGFIX: change Labeled types labels from lists to sets.
+
 1.9.4 (2019-10-16)
 ------------------
 
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -303,8 +303,23 @@ class Uuid(Metric):
 
 @dataclass
 class Labeled(Metric):
-    labels: List[str] = None
     labeled = True
+    # Start the labels as a List to preserve the order from `metrics.yaml`
+    labels: List[str] = None
+
+    def __post_init__(self, *args):
+        ordered_labels = self.labels
+
+        # If any, turn the labels into a Set to properly inialize the Metric object.
+        # This is important because Kotlin code needs this labels as a Set
+        if self.labels is not None:
+            self.labels = set(ordered_labels)
+        super().__post_init__(*args)
+
+        # After initializing the Metric object, add the ordered_labels to it.
+        # If we do this step before calling the __post_init__ for the Metric class,
+        # the metric schema validation will fail
+        self.ordered_labels = ordered_labels
 
 
 @dataclass
--- a/glean_parser/markdown.py
+++ b/glean_parser/markdown.py
@@ -25,8 +25,8 @@ def extra_info(obj):
         for key in obj.allowed_extra_keys:
             extra_info.append((key, obj.extra_keys[key]["description"]))
 
-    if isinstance(obj, metrics.Labeled) and obj.labels is not None:
-        for label in obj.labels:
+    if isinstance(obj, metrics.Labeled) and obj.ordered_labels is not None:
+        for label in obj.ordered_labels:
             extra_info.append((label, None))
 
     return extra_info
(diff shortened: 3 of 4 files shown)


brizental:
I made sure this time that the Kotlin and Swift generated codes were not changed ;)

brizental:
I am looking into why this failed. I ran the `make test` command locally and it was successful, but the one that is failing is `make test-all`.

HISTORY.rst
```diff
@@ -5,6 +5,11 @@ History
 Unreleased
 ----------
 
+1.9.5 (2019-10-18)
```
badboy: Leave this line out, we should add this once we actually release a version, for now stuff should be kept under the unreleased label

mdboom:
> I am looking into why this failed. I ran the make test command locally and it was successful, but the one that is failing is make test-all.

Sometimes with Python it's difficult to ensure that you are testing your local changes and not some previously installed version.  To confirm -- did you install with `python setup.py develop` which just installs symlinks so that local changes in your checkout are automatically reflected.  And make sure that the python interpreter you are testing with is the same you are installing with.  Sometimes `py.test` might use a different Python interpreter and you could try 

```
python `which py.test`
```

brizental:
Okay, I was able to get this to work. 

I am not happy with the solution since it looks totally hacky, but it works so I am putting it up for review. If this gets accepted I can a put a comment there just so that it is understandable. I would  be up for changing it completely, if someone has a better idea on how to approach this.

glean_parser/metrics.py
```diff
@@ -306,6 +306,13 @@ class Labeled(Metric):
     labels: List[str] = None
     labeled = True
```
brizental: One thing that might be worrying: during development I accidentally left this value as `False` and all the testsstill passed... Maybe there should some more specific tests that would prevent that from being the case.

mdboom: Yeah -- that's a good idea.

HISTORY.rst
badboy: True, but there are now more changes (so requires a rebase anyway) plus the date is not correct now.

glean_parser/metrics.py
```diff
@@ -306,6 +306,13 @@ class Labeled(Metric):
...
 
+    def __post_init__(self, *args):
+        ordered_labels = self.labels
```
Dexterp37: Instead of doing this, one other option could be to check if the metric type is a LabeledMetricType and do the fix-up [here](https://github.com/mozilla/glean_parser/blob/83d3af92354569cb9c87a1b1851f342397c4a234/glean_parser/kotlin.py#L19). @mdboom , do you have opinions on any of the two solutions?

mdboom: I think this is cleaner, and it will apply to all output formats.

Dexterp37: Then @brizental let's add a comment here explaining what is this for, please :)

brizental: In this context, does `this` mean my fix os Alessio's suggestion?

mdboom: Sorry.  I think @brizental's fix using the `__post_init__` method is cleaner.

HISTORY.rst
brizental: Makes sense, I will just move that to `Unreleased`.

HISTORY.rst
```diff
@@ -9,7 +9,7 @@ Unreleased
 
 * New lint: Restrict what metric can go into the 'baseline' ping
 
-* New lint: Warn for slight misspellings in ping names
```
mdboom: I think maybe this was a merge conflict resolved incorrectly.  I don't think we want to remove this line (just add the new one below).

brizental: Hehe, oops. Fixed :)

PR #129, opened 2019-10-21
Ignore lower-camelcase class name and enum variants in linter

badboy:
**DO NOT MERGE UNTIL [BUG 1590080](https://bugzilla.mozilla.org/show_bug.cgi?id=1590080) REACHED A CONCLUSION**

This might be a short-term fix until we decide that `ClickKeys` and
`Open` are better than `clickKeys` and `open`, but we will need to fix
these things in Fenix when upgrading.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,8 @@ History
 Unreleased
 ----------
 
+*  Ignore lower-camelcase class name and enum variants in linter
+
 1.9.4 (2019-10-16)
 ------------------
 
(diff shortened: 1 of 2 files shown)


badboy:
**DO NOT MERGE UNTIL [BUG 1590080](https://bugzilla.mozilla.org/show_bug.cgi?id=1590080) REACHED A CONCLUSION**

mdboom:
We currently run `ktlint` (but not `detekt`) over the generated code.  It would be nice to have a unit test that generated code with enums in it and ran it through `detekt` so we catch this sort of stuff earlier.

Dexterp37:
> 
> 
> This might be a short-term fix until we decide that `ClickKeys` and
> `Open` are better than `clickKeys` and `open`, but we will need to fix
> these things in Fenix when upgrading.

We can discuss the long term solution in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1590080). Since no repo is failing due to this, I'd personally hold it back: it might make more sense to generate code that better suites the Kotlin guidelines for Android.

PR #131, opened 2019-10-22
Run detekt over generated Kotlin code in unit tests

mdboom:
This sets the stage for, but doesn't really fix [1590080](https://bugzilla.mozilla.org/show_bug.cgi?id=1590080).  This just adds `detekt` to the checks performed on the generated Kotlin code in CI, and adds the necessary supressions to make them pass.

### DIFF
--- a/glean_parser/templates/kotlin.jinja2
+++ b/glean_parser/templates/kotlin.jinja2
@@ -19,7 +19,7 @@ Jinja2 template is not. Please file bugs! #}
         )
 {% if lazy %}    }{% endif %}{% endmacro %}
 
-@file:Suppress("PackageNaming")
+@file:Suppress("PackageNaming", "MaxLineLength")
 package {{ namespace }}
 
 import {{ glean_namespace }}.private.HistogramType // ktlint-disable import-ordering no-unused-imports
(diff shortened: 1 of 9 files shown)


glean_parser/templates/kotlin.jinja2
```diff
@@ -19,7 +19,7 @@ Jinja2 template is not. Please file bugs! #}
         )
 {% if lazy %}    }{% endif %}{% endmacro %}
 
-@file:Suppress("PackageNaming")
+@file:Suppress("PackageNaming", "MaxLineLength")
```
mdboom: This is triggered when a labeled metric has lots of labels and the `listOf(...)` construction is really long.  Fixing that properly would require a tool that can properly re-indent Kotlin code.  `ktlint` can do that, but that would make `ktlint` a hard requirement of glean_parser, which is currently isn't.  (ktlint is only used for unit tests right now).

glean_parser/templates/kotlin.jinja2
```diff
@@ -37,6 +37,7 @@ import {{ glean_namespace }}.private.LabeledMetricType // ktlint-disable import-
 internal object {{ category_name|Camelize }} {
 {% for obj in objs.values() %}
     {% if obj.extra_keys|length %}
+    @Suppress("ClassNaming", "EnumNaming")
```
mdboom: Just suppressing these now because fixing would require a change in the public API.

glean_parser/templates/kotlin.jinja2
Dexterp37: Good catch, no need to depend on ktlint for this..

glean_parser/templates/kotlin.jinja2
Dexterp37: Do we have a bug for fixing this? If so, can you mention it as a comment in here?

mdboom: I thought we could just keep [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1590080) to track that.  I don't consider this PR as fixing that bug, just making it easier to address that bug properly when we decide how to proceed (and preventing bugs of that type in the future)...

PR #132, opened 2019-10-22
Test in Python 3.8

mdboom:


HISTORY.rst
```diff
@@ -5,6 +5,10 @@ History
 Unreleased
 ----------
 
+* Python 3.8 is now tested in CI (in addition to Python 3.7).
```
Dexterp37: Eugh, well, tox wasn't working too much for me as well :) So I won't be sad about this!

mdboom: Yeah, the tl;dr is that CircleCI version 1 provided a way to install a bunch of different Python versions in the same container/VM which is sort of the main use case for `tox`.  The new way is to use a separate container for each Python version.  It does slightly more work that way, but it's probably better from an isolation perspective.

PR #143, opened 2019-11-30
WIP: 1600257: Support Python 3.5

mdboom:
This PR makes me :cry:

Basically, the changes are:
- Don't use `dataclasses`.  Most of the type checking it provided for us was redundant anyway, since we validate using the JSON schema before getting there anyway.
- `pathlib.Path` didn't play well with the rest of the `stdlib` in Python 3.5, so we need to do more explicit casting to strings.
- `datetime.datetime.fromisoformat()` is new in 3.7, so this uses a backport library
- `__init_subclass__` is new in 3.6, so we use the `pep487` backport library for that
- f-strings are new in 3.6, so we have to go back to `str.format()` :nauseated_face: 
- Dictionaries have a stable ordering in 3.7, so here we have to use `OrderedDict` in a number of places to make the tests pass consistently.

And, of course, this also adds CI for Python 3.5 and 3.6.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,11 +29,6 @@ commands:
             - run:
                     name: test
                     command: make test
-            - run:
-                    name: make-docs
-                    command: |
-                        make docs
-                        touch docs/_build/html/.nojekyll
 
 jobs:
     build-35:
@@ -57,6 +52,11 @@ jobs:
             - persist_to_workspace:
                     root: docs/_build
                     paths: html
+            - run:
+                    name: make-docs
+                    command: |
+                        make docs
+                        touch docs/_build/html/.nojekyll
 
     build-38:
         docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 commands:
     test-python-version:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ jobs:
             - image: circleci/python:3.7.5
         steps:
             - test-python-version
-            - persist-to-workspace:
+            - persist_to_workspace:
                     root: docs/_build
                     paths: html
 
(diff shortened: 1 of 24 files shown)


.circleci/config.yml
```diff
@@ -1,49 +1,68 @@
...
+                    name: install
+                    command: python setup.py install --user
+            - run:
+                    name: test
+                    command: make test
 
 jobs:
+    build-35:
+        docker:
+            - image: circleci/python:3.5.9
```
Dexterp37: The version used on Mozilla build machines is [Python 3.5.3](https://bugzilla.mozilla.org/show_bug.cgi?id=1598350#c5). Does it make sense to use that specific image?

mdboom: Python is generally pretty good about not breaking things with micro releases, but, sure, it probably makes sense to use the exact version for good measure.

mdboom:
As discussed in person, let's hold off on merging this until we've had a chance to explore the (low) possibility of upgrading Python in m-c.

Dexterp37:
@mdboom we can go ahead and merge this, [see this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1598350#c12).

.circleci/config.yml
mdboom: It turns out it was kind of nightmare to test against 3.5.3.  The 3.5.3 CircleCI Docker image has an entirely different (older) version of Debian, so the version of Java available is too old to run ktlint and detekt, and the version of `pip` is too old to install dependencies from a file.  I'm quite confident in Python that there weren't any breaking changes between 3.5.3 and 3.5.9, and it's easier to keep the rest of the non-Python part of our CI the same across all Python versions.

So I think we keep this as-is on 3.5.9 and re-address this if we run into issues trying to use it in m-c (which we should know immediately before merging into the m-c tree so is pretty low risk).

Dexterp37: Sounds good to me!

PR #146, opened 2019-12-11
Bug 1601270 Restrict ping names to be kebab-case

jklukas:
This supersedes #90 and #94, and adds special cases to
allow current legacy ping names with underscores to continue
to work.

### DIFF
--- a/glean_parser/schemas/metrics.1-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.1-0-0.schema.yaml
@@ -28,6 +28,13 @@ definitions:
     pattern: "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})*$"
     maxLength: 40
 
+  kebab_case:
+    type: string
+    # Bug 1601270; we allow 3 specific existing snake_cased ping names for now,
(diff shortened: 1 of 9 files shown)


glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -28,6 +28,13 @@ definitions:
     pattern: "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})*$"
     maxLength: 40
 
+  kebab_case:
+    type: string
+    # Bug 1601270; we allow 3 specific existing snake_cased ping names for now,
+    # but these special cases can be removed in 2020H2.
```
Dexterp37: Ah, why can these be removed in 2020H2?

jklukas: I'm assuming the number of clients sending pings with the snake case names will have died down to negligible levels by then, but I also plan to add monitoring in the data pipeline to tell us how often we're seeing snake_cased ping names, so that we can understand the impact if we remove support.

I'll add more to the comment here.

PR #148, opened 2019-12-12
Use iso8601 library rather than backports-fromisoformat

mdboom:
The reasons for this PR are a bit obscure.

We were using the `backports.datetime_fromisoformat` library to support parsing iso8601 dates on Python 3.6 and earlier (where such functionality is not in the standard library).  However, this library includes a C extension, so installing it on Windows requires a C compiler to be handy and installed correctly so Python can find it (which is not always the case, even if the user otherwise has their machine set up for C/C++ development).

This replaces that library with another one that is pure Python to avoid this issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,7 +33,7 @@ is installed by `pip`.
 
 Additionally on Python 3.6 and 3.5:
 
-- backports-datetime-fromisoformat
+- iso8601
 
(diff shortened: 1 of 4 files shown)


mdboom:
Making an executive decision to merge this without review, because it has broken the a-c build.

PR #152, opened 2019-12-13
Break the CI build if any of the dependencies require a C compiler

mdboom:
This should prevent against future errors like #148 where we inadvertently added a dependency that requires a C compiler at install time.  We can't count on a working C compiler to be available every where that glean_parser is used, so we should detect when we use a dependency like that and break the build.

(NOTE: I expect this PR to fail the first time as a way of testing it, and then will come back and remove the "bad" dependency to make it pass.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,13 @@ commands:
                     command: make lint
             - run:
                     name: install
-                    command: python setup.py install --user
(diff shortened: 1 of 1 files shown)


mdboom:
A CircleCI build showing that this breaks things when we want it to: https://circleci.com/gh/mozilla/glean_parser/936

PR #154, opened 2019-12-19
Document pings.yaml schema

badboy:

### DIFF
--- a/docs/sphinxext/eval.py
+++ b/docs/sphinxext/eval.py
@@ -4,6 +4,7 @@ from docutils import statemachine
 import m2r
 
 from glean_parser.parser import get_parameter_doc
+from glean_parser.parser import get_ping_parameter_doc
 
 
 class MetricParameterDirective(Directive):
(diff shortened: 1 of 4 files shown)


docs/sphinxext/eval.py
```diff
@@ -36,5 +37,36 @@ def run(self):
         return []
 
 
+class PingParameterDirective(Directive):
```
mdboom: Can we reduce the duplication here?  It looks like the only difference is the `get_(ping)?parameter_doc` part.  Maybe make that a method overridden in each class that inherits from a common base class?

glean_parser/parser.py
```diff
@@ -128,6 +128,15 @@ def get_parameter_doc(key):
     return schema["definitions"]["metric"]["properties"][key]["description"]
 
 
+def get_ping_parameter_doc(key):
+    """
+    Returns documentation about a specific ping parameter.
+    """
+    schema, _ = _get_schema(PINGS_ID)
+    print(str(schema))
```
mdboom: Debugging print?

PR #158, opened 2020-01-17
Update ping schema url

brizental:
I have updated the schema and changed it to look exactly like it does in glean-core (see: https://github.com/mozilla/glean/blob/master/glean-core/android/build.gradle#L26-L27).

This is one case where it would be useful to have shared global vars between the projects. That is not urgent of course, I have closed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1551694) related to this last year because the changes were too much work and not much gain, but still, just to keep it in mind :)

Ah, also did a small change to the Makefile. A bunch of my tests failed when I updated the schema, turns out it was not related to that, but to the fact that the project was not building because it couldn't get past the `clean` phase 🤦‍♀️

### DIFF
--- a/glean_parser/validate_ping.py
+++ b/glean_parser/validate_ping.py
@@ -26,7 +26,7 @@ SCHEMAS_DIR = ROOT_DIR / "schemas"
 GLEAN_PING_SCHEMA_GIT_HASH = "a56043b"
 GLEAN_PING_SCHEMA_URL = (
     "https://raw.githubusercontent.com/mozilla-services/"
-    "mozilla-pipeline-schemas/%s/schemas/glean/glean/"
+    "mozilla-pipeline-schemas/{}/schemas/glean/glean/"
     "glean.1.schema.json"
 ).format(GLEAN_PING_SCHEMA_GIT_HASH)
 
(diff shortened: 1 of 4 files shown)


glean_parser/validate_ping.py
```diff
@@ -23,17 +23,18 @@
...
 
 
-PING_SCHEMA_DEFAULT_URL = (
+GLEAN_PING_SCHEMA_GIT_HASH = "a56043b"
+GLEAN_PING_SCHEMA_URL = (
     "https://raw.githubusercontent.com/mozilla-services/"
-    "mozilla-pipeline-schemas/dev/schemas/glean/baseline/"
-    "baseline.1.schema.json"
-)
+    "mozilla-pipeline-schemas/%s/schemas/glean/glean/"
```
mdboom:
```suggestion
    "mozilla-pipeline-schemas/{}/schemas/glean/glean/"
```

`.format` uses the newer format syntax.  The `%s` only works with the deprecated `%` operator.

brizental:
> The one that matters is here, because we want the version of the schema to test against to track the development of Glean itself.

That's true, although changing that hash does not change the python tests in glean-core only the android ones. The python tests are relying on the default one that I changed here.

Thinking about this now I could've just fixed this by adding the `schema_url` argument to the `validate_ping` calls in glean-core. Anyways.

Thanks for the quick review!

brizental:
After discussion in IRC we decided to remove completely ths default, so enforce that the schema_url is always set. I will close this PR and open another one with the new solution.

mdboom:
> That's true, although changing that hash does not change the python tests in glean-core only the android ones. The python tests are relying on the default URL that I changed here.

Yes, thanks for noticing that!  That was never the intention.

PR #159, opened 2020-01-17
Make 'schema_url' required for 'validate_ping'

brizental:

### DIFF
--- a/glean_parser/__main__.py
+++ b/glean_parser/__main__.py
@@ -74,9 +74,8 @@ def translate(input, format, output, option, allow_reserved):
     "--schema",
     "-s",
     type=str,
-    default=validate_ping.PING_SCHEMA_DEFAULT_URL,
     nargs=1,
-    required=False,
+    required=True,
     help=("HTTP url or file path to Glean ping schema. If remote, will cache to disk."),
 )
 def check(schema):
--- a/glean_parser/validate_ping.py
+++ b/glean_parser/validate_ping.py
@@ -23,18 +23,9 @@ ROOT_DIR = Path(__file__).parent
 SCHEMAS_DIR = ROOT_DIR / "schemas"
 
 
-PING_SCHEMA_DEFAULT_URL = (
-    "https://raw.githubusercontent.com/mozilla-services/"
-    "mozilla-pipeline-schemas/dev/schemas/glean/baseline/"
-    "baseline.1.schema.json"
-)
-
-
 @functools.lru_cache(maxsize=1)
 def _get_ping_schema(schema_url):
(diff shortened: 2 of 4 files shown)


glean_parser/__main__.py
```diff
@@ -74,7 +80,7 @@ def translate(input, format, output, option, allow_reserved):
     "--schema",
     "-s",
     type=str,
-    default=validate_ping.PING_SCHEMA_DEFAULT_URL,
+    default=PING_SCHEMA_DEFAULT_URL,
     nargs=1,
     required=False,
```
mdboom: I think we can also make this `required=True` (and remove the default).  That will change the CLI, so we'll have to confirm it doesn't break things, but we should be able to know when we file a PR to update `glean_parser` in `glean` before merging it, so low risk...

glean_parser/util.py
```diff
@@ -219,10 +217,9 @@ def resolve_remote(self, uri):
     return NullResolver.from_schema(schema)
 
 
-def fetch_remote_url(url, cache=True):
```
mdboom: I think we still want to cache.  This is important because caching means we don't have to be connected to the internet every time we run Glean unit tests.  Was there something that broke by not doing this?

glean_parser/validate_ping.py
```diff
@@ -23,18 +23,9 @@
...
-    "baseline.1.schema.json"
-)
-
-
 @functools.lru_cache(maxsize=1)
 def _get_ping_schema(schema_url):
-    contents = util.fetch_remote_url(
-        schema_url, cache=(schema_url != PING_SCHEMA_DEFAULT_URL)
-    )
+    contents = util.fetch_remote_url(schema_url)
```
mdboom: We could just have `cache=True` here (and reinstate the caching behavior above).

glean_parser/validate_ping.py
```diff
@@ -59,17 +50,18 @@ def _validate_ping(ins, outs, schema_url):
...
 
     :param ins: Input stream or file path to the ping contents to validate
     :param outs: Output stream to write errors to. (Defaults to stdout)
     :param schema_url: HTTP URL or local filesystem path to Glean ping schema.
-        Defaults to the current version of the schema in
-        mozilla-pipeline-schemas.
     :rtype: int 1 if any errors occurred, otherwise 0.
     """
+    if schema_url is None:
+        raise ValueError("Missing required argument 'schema_url'")
```
mdboom: Missing arguments are usually `TypeError` in Python (not sure why, it doesn't make much sense, but let's be consistent with that)...

PR #161, opened 2020-01-23
1611111 - Restore the order of extra arguments in Swift code

badboy:
This fixes 2 small bugs (one is a drive-by fix).

Caution: mozilla/glean will still not compile with this, due to the reason code changes, that don't exist yet for Swift.

### DIFF
--- a/glean_parser/swift.py
+++ b/glean_parser/swift.py
@@ -121,16 +121,18 @@ def output_swift(objs, output_dir, options={}):
     )
 
     # The object parameters to pass to constructors.
-    # Need to be in the order the type constructor expects them.
+    # **CAUTION**: This list needs to be in the order the type constructor expects them.
+    # The `test_order_of_fields` test checks that the generated code is valid.
+    # **DO NOT CHANGE THE ORDER OR ADD NEW FIELDS IN THE MIDDLE**
     extra_args = [
-        "allowed_extra_keys",
         "category",
-        "disabled",
-        "lifetime",
         "name",
-        "reason_codes",
         "send_in_pings",
+        "lifetime",
+        "disabled",
         "time_unit",
+        "reason_codes",
+        "allowed_extra_keys",
     ]
 
     namespace = options.get("namespace", "GleanMetrics")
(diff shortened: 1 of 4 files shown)


glean_parser/swift.py
```diff
@@ -121,16 +121,18 @@ def output_swift(objs, output_dir, options={}):
...
         "category",
-        "disabled",
-        "lifetime",
         "name",
-        "reason_codes",
         "send_in_pings",
+        "lifetime",
+        "disabled",
         "time_unit",
+        "reason_codes",
```
mdboom: Why don't we just take this out here.  We can put it back in right before https://github.com/mozilla/glean/pull/649 is ready to land.

glean_parser/templates/swift.jinja2
```diff
@@ -15,7 +15,7 @@ Jinja2 template is not. Please file bugs! #}
         )
 {% endmacro %}
 
-{% if glean_namespace|length and glean_namespace != "Glean" %}
+{% if glean_namespace|length and glean_namespace is not equalto "Glean" %}
```
mdboom: Jinja is weird.

badboy: It is! In all my _local_ and manual tests it worked just fine with `!=`, but not in Glean aparently ¯\_(ツ)_/¯

glean_parser/swift.py
badboy: Need to test it, that might allow us to land the Swift part of reason codes separately.

badboy: The real problem with reason codes is currently what code is generated:

```swift
let customPing = Ping<NoExtraKeys>(...)
```

We might just want to skip that `<NoExtraKeys>` part for Swift right now or we need to make glean#649 include the Swift part.

mdboom: Yeah -- I'm working on the Swift part for #649 -- the problem is just what we should land in glean_parser now to make sure glean master is working.

badboy: Glean master is still on an old parser version, which works. It currently won't work with 1.16.0

mdboom: Oh -- my bad.  I thought we needed to upgrade glean_parser in glean for other reasons.  If not, let's not worry -- we can but a new glean_parser release and update to it as part of #649.

badboy: 👍 Sounds good to me.

PR #174, opened 2020-02-21
Drop 'inflection' and make transitive dependencies versions explicit

Dexterp37:
This is required in order to integrate in mozilla-central, [see this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1602773).

Given :ahal comment on the bug, I checked about downgrading pyYAML to 3.11. However 3.13 contains [fixes for python 3.7](https://github.com/yaml/pyyaml/blob/2f463cf5b0e98a52bc20e348d1e69761bf263b86/CHANGES#L83), which we need to support. I'll attempt to update pyyaml in m-c.

### DIFF
--- a/glean_parser/util.py
+++ b/glean_parser/util.py
@@ -15,7 +15,6 @@ import urllib.request
 
 import appdirs
 import diskcache
-import inflection
 import jinja2
 import jsonschema
 from jsonschema import _utils
@@ -136,6 +135,24 @@ def ensure_list(value):
     return value
 
 
+def to_camel_case(input, capitalize_first_letter):
+    """
+    Convert the value to camelCase.
+
+    This additionally replaces any '.' with '_'. The first letter is capitalized
+    depending on `capitalize_first_letter`.
+    """
+    sanitized_input = input.replace(".", "_").replace("-", "_")
+    # Filter out any empty token. This could happen due to leading '_' or
+    # consecutive '__'.
+    tokens = [s.capitalize() for s in sanitized_input.split("_") if len(s) != 0]
+    # If we're not meant to capitalize the first letter, then lowercase it.
+    if not capitalize_first_letter:
+        tokens[0] = tokens[0].lower()
+    # Finally join the tokens and capitalize.
+    return ''.join(tokens)
+
+
 def camelize(value):
     """
     Convert the value to camelCase (with a lower case first letter).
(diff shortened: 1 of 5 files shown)


glean_parser/util.py
```diff
@@ -136,14 +135,32 @@ def ensure_list(value):
...
+    """
+    sanitized_input = input.replace(".", "_").replace("-", "_")
+    # Filter out any empty token. This could happen due to leading '_' or
+    # consecutive '__'.
+    tokens = [s.capitalize() for s in sanitized_input.split("_") if len(s) != 0]
+    # If we're not meant to capitalize the first letter, then lowercase it.
+    if not capitalize_first_letter:
+        tokens[0] = tokens[0].lower()
+    # Finally join the tokens and capitalize.
+    return ''.join(tokens)
```
mdboom: We could, alternatively, take the regex approach used in inflection, which given how Python works is probably much faster (which would require adding a copyright notice about the borrowing of code, of course).

setup.py
```diff
@@ -27,13 +27,18 @@
     "appdirs>=1.4.3",
     "Click>=7.0",
     "diskcache>=4.0.0",
-    "inflection>=0.3.1",
     "iso8601>=0.1.12",
     "Jinja2>=2.10.1,<3.0",
     "jsonschema>=3.0.2",
+    # 'markupsafe' is required by Jinja2. From version 2.0.0 on
+    # py3.5 support is dropped.
+    "markupsafe<2.0.0",
```
mdboom: Just to be safe, lets add `markupsafe>=1.1,<2.0.0` so we specify the minimum version of this required by the version of Jinja2 we are using.  The resolver *should* work that out on its own, but in my experience `pip` doesn't always do a great job of that, so more hints is usually better.

glean_parser/util.py
Dexterp37: I thought about doing that, but I opted for rewriting this given my poor knowledge of regexes and adding more specific test cases for our purpose. I wasn't really sure how to handle copyright as well :(

Moreover, the 'inflection' library does not properly handle strings such as `_test__me`, which are transformed to `_test_me`. I'm not sure how often we'd hit that case (if any time at all!), but if you have no strong opinion on this I'd leave it as it is if only to save us the copyright trouble :)

setup.py
Dexterp37: Great point!

PR #176, opened 2020-02-25
Add more information to pings md docs

brizental:
Fixes [Bug 1615221](https://bugzilla.mozilla.org/show_bug.cgi?id=1615221)

* Add info about wether the ping includes client id;
* Add list of data review links;
* Add list of related bugs links.

### DIFF
--- a/glean_parser/templates/markdown.jinja2
+++ b/glean_parser/templates/markdown.jinja2
@@ -29,8 +29,26 @@ See the Glean SDK documentation for the [`{{ ping_name }}` ping]({{ ping_name|pi
 This ping is sent if empty.
 
 {% endif %}
+{% if ping_name|ping_include_client_id %}
+This ping includes the client id.
+
+{% endif %}
+{% if ping_name|ping_data_reviews %}
+**Data reviews for this ping:**
+{% for review in ping_name|ping_data_reviews %}
+    - [{{review}}]({{review}})
+{% endfor %}
+
+{% endif %}
+{% if ping_name|ping_bugs %}
+**Bugs related to this ping:**
+{% for bug in ping_name|ping_bugs %}
+    - [Bug {{bug}}](https://bugzilla.mozilla.org/show_bug.cgi?id={{bug}})
+{% endfor %}
+
+{% endif %}
 {% if ping_name|ping_reasons %}
-Reasons this ping may be sent:
+**Reasons this ping may be sent:**
 
 {% for (reason, desc) in ping_name|ping_reasons|dictsort %}
     - `{{ reason }}`: {{ desc|indent(6, indentfirst=False) }}
--- a/glean_parser/templates/markdown.jinja2
+++ b/glean_parser/templates/markdown.jinja2
@@ -30,7 +30,7 @@ This ping is sent if empty.
 
 {% endif %}
 {% if ping_name|ping_include_client_id %}
-This ping includes the [client id](https://mozilla.github.io/glean/book/user/pings/index.html?highlight=client_info#the-client_info-section).
+This ping includes the [client id](https://mozilla.github.io/glean/book/user/pings/index.html#the-client_info-section).
 
 {% endif %}
 {% if ping_name|ping_data_reviews %}
--- a/glean_parser/templates/markdown.jinja2
+++ b/glean_parser/templates/markdown.jinja2
@@ -30,20 +30,23 @@ This ping is sent if empty.
 
 {% endif %}
 {% if ping_name|ping_include_client_id %}
-This ping includes the client id.
+This ping includes the [client id](https://mozilla.github.io/glean/book/user/pings/index.html?highlight=client_info#the-client_info-section).
 
 {% endif %}
 {% if ping_name|ping_data_reviews %}
 **Data reviews for this ping:**
+
 {% for review in ping_name|ping_data_reviews %}
-    - [{{review}}]({{review}})
+    - <{{review}}>
 {% endfor %}
 
 {% endif %}
 {% if ping_name|ping_bugs %}
 **Bugs related to this ping:**
+
 {% for bug in ping_name|ping_bugs %}
-    - [Bug {{bug}}](https://bugzilla.mozilla.org/show_bug.cgi?id={{bug}})
+    - {% if bug|int != 0 %}{{bug}}{% else %}<{{bug}}>{% endif %}
+
 {% endfor %}
 
 {% endif %}
--- a/glean_parser/templates/markdown.jinja2
+++ b/glean_parser/templates/markdown.jinja2
@@ -37,7 +37,7 @@ This ping includes the [client id](https://mozilla.github.io/glean/book/user/pin
 **Data reviews for this ping:**
 
 {% for review in ping_name|ping_data_reviews %}
-    - <{{review}}>
+- <{{review}}>
 {% endfor %}
 
 {% endif %}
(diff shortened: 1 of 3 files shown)


glean_parser/templates/markdown.jinja2
```diff
@@ -28,9 +28,27 @@ See the Glean SDK documentation for the [`{{ ping_name }}` ping]({{ ping_name|pi
...
 
+{% endif %}
+{% if ping_name|ping_include_client_id %}
+This ping includes the client id.
+
+{% endif %}
+{% if ping_name|ping_data_reviews %}
+**Data reviews for this ping:**
+{% for review in ping_name|ping_data_reviews %}
+    - [{{review}}]({{review}})
```
badboy: If `review` is an actual URL markdown has a short variant for that: `<url>`

glean_parser/templates/markdown.jinja2
```diff
@@ -28,9 +28,27 @@ See the Glean SDK documentation for the [`{{ ping_name }}` ping]({{ ping_name|pi
...
+{% endfor %}
+
+{% endif %}
+{% if ping_name|ping_bugs %}
+**Bugs related to this ping:**
+{% for bug in ping_name|ping_bugs %}
+    - [Bug {{bug}}](https://bugzilla.mozilla.org/show_bug.cgi?id={{bug}})
```
badboy: Afaik we now enforce these to be URLs regardless, as they could also point to GitHub or other places.

glean_parser/templates/markdown.jinja2
brizental: Hm, documentation says these shoud be URIs. So I guess there is no guarantee they can be linked.

From https://mozilla.github.io/glean_parser/pings-yaml.html:
> A list of URIs to any data collection reviews relevant to the ping.

glean_parser/templates/markdown.jinja2
brizental: Also from https://mozilla.github.io/glean_parser/pings-yaml.html:
> If a number, it is an ID to an issue in the default tracker (e.g. Mozilla’s Bugzilla instance). If a string, it must be a URI to a bug page in a tracker.

So I guess I should verify before creating the link. I am guessing this is just for backwars compatibility from when these used to be only numbers?

glean_parser/templates/markdown.jinja2
badboy: These _are_ URIs, so above is fine, we can use them as link.
I just offered a different way of expressing that.

glean_parser/templates/markdown.jinja2
badboy: Indeed, we still do allow integers, so we would need to check indeed.

chutten: Is that only for backwards compatibility? I thought we'd made the decision to permit URLs only

glean_parser/templates/markdown.jinja2
```diff
@@ -28,9 +28,27 @@ See the Glean SDK documentation for the [`{{ ping_name }}` ping]({{ ping_name|pi
 {% if ping_name|ping_send_if_empty %}
 This ping is sent if empty.
 
+{% endif %}
+{% if ping_name|ping_include_client_id %}
+This ping includes the client id.
```
chutten: Maybe link "client id" to the `client_info` documentation in the Book of Glean?

glean_parser/templates/markdown.jinja2
brizental: My point is that the documentation states URI, which doesn't necessarily mean URL, which means it might not be linkable at all.

glean_parser/templates/markdown.jinja2
badboy: I thought so too!
I think we _did_ change all IDs to URLs in projects we can influence (citation needed) and [the glinter suggests that it's deprecated](https://github.com/mozilla/glean_parser/blob/master/glean_parser/lint.py#L142-L149).
However we didn't change the schema because we need to parse old stuff (as in: the probe-scraper parses)

glean_parser/templates/markdown.jinja2
brizental: Probably overthinking this, hehe. I do prefer the <url> syntax and didn't know it before :) I'll change it.

glean_parser/templates/markdown.jinja2
```diff
@@ -28,9 +28,30 @@ See the Glean SDK documentation for the [`{{ ping_name }}` ping]({{ ping_name|pi
 {% if ping_name|ping_send_if_empty %}
 This ping is sent if empty.
 
+{% endif %}
+{% if ping_name|ping_include_client_id %}
+This ping includes the [client id](https://mozilla.github.io/glean/book/user/pings/index.html?highlight=client_info#the-client_info-section).
```
chutten: Not sure we want to highlight `client_info`:
```suggestion
This ping includes the [client id](https://mozilla.github.io/glean/book/user/pings/index.html#the-client_info-section).
```

glean_parser/templates/markdown.jinja2
```diff
@@ -28,9 +28,30 @@ See the Glean SDK documentation for the [`{{ ping_name }}` ping]({{ ping_name|pi
...
+{% for review in ping_name|ping_data_reviews %}
+    - <{{review}}>
+{% endfor %}
+
+{% endif %}
+{% if ping_name|ping_bugs %}
+**Bugs related to this ping:**
+
+{% for bug in ping_name|ping_bugs %}
+    - {% if bug|int != 0 %}{{bug}}{% else %}<{{bug}}>{% endif %}
```
chutten: Linking the bug number to BMO would be the best usability here... but on the other hand, let's not make bug numbers (a thing we deprecated) any nicer than they have to be : )

Dexterp37:
@brizental can you kindly provide a sample rendered page?

glean_parser/templates/markdown.jinja2
brizental: My decision to not do that was because it is not stated that if this is a number, it will be a Bugzilla bug number.

brizental:
For a `pings.yaml` that looks like this: 

```yaml
# Any copyright is dedicated to the Public Domain.
# https://creativecommons.org/publicdomain/zero/1.0/

---
$schema: moz://mozilla.org/schemas/glean/pings/1-0-0

custom-ping-might-be-empty:
  description:
    This is another custom ping
  include_client_id: true
  send_if_empty: true
  bugs:
    - 1137353
    - http://nowhere.com/reviews
  data_reviews:
    - http://nowhere.com/reviews
  notification_emails:
    - CHANGE-ME@test-only.com
  reasons:
    silly:
      A silly reason for sending a ping.
    serious:
      A serious reason for sending a ping.
```

The rendered page will look like this:



# Metrics
This document enumerates the metrics collected by this project.
This project may depend on other projects which also collect metrics.
This means you might have to go searching through the dependency tree to get a full picture of everything collected by this project.

# Pings

 - [custom-ping-might-be-empty](#custom-ping-might-be-empty)


## custom-ping-might-be-empty

This is another custom ping

This ping is sent if empty.

This ping includes the [client id](https://mozilla.github.io/glean/book/user/pings/index.html?highlight=client_info#the-client_info-section).

**Data reviews for this ping:**

- <http://nowhere.com/reviews>

**Bugs related to this ping:**

- 1137353
- <http://nowhere.com/reviews>

**Reasons this ping may be sent:**

- `serious`: A serious reason for sending a ping.
- `silly`: A silly reason for sending a ping.

This ping contains no metrics.

glean_parser/templates/markdown.jinja2
Dexterp37: > 
> 
> My decision to not do that was because it is not stated that if this is a number, it will be a Bugzilla bug number.

Yes, and that's a good call. Fenix used to have issue numbers there, that's why we moved to requiring URIs.

PR #177, opened 2020-02-28
1618873: Support external translators

mdboom:
This is just a minor refactor to create an entry point where another glean_parser-using-library can provide their own translation function.  It doesn't provide a *ton* of value over what could be done without this change except providing access to the transactional directory writing behavior.

### DIFF
--- a/glean_parser/translate.py
+++ b/glean_parser/translate.py
@@ -24,59 +24,62 @@ from . import swift
 from . import util
 
 
-# Each outputter in the table has the following keys:
-# - "output_func": the main function of the outputter, the one which
-#   does the actual translation.
-# - "clear_output_dir": a flag to clear the target directory before moving there
-#   the generated files.
-# - "extensions": A list of glob patterns to clear in the directory.
-
-
 class Outputter:
     """
     Class to define an output format.
+
+    Each outputter in the table has the following member values:
+    - output_func: the main function of the outputter, the one which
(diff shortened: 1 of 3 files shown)


glean_parser/translate.py
```diff
@@ -25,39 +25,44 @@
 # Each outputter in the table has the following keys:
 # - "output_func": the main function of the outputter, the one which
 #   does the actual translation.
-# - "clear_output_dir": a flag to clear the target directory before moving there
-#   the generated files.
+# - "extensions": A list of glob patterns on files to clear before generating
```
Dexterp37: nit: can we rename this to something that hints at file deletion? Like `clear_pattern` (not the best choice) or something similar?

tests/test_translate.py
```diff
@@ -135,3 +136,29 @@ def test_translate_dont_remove_extra_files(tmpdir):
...
+
+    def external_translator(all_objects, output_dir, options):
+        assert {"foo": "bar", "allow_reserved": True} == options
+
+        for category in all_objects:
+            with (tmpdir / category).open('w') as fd:
+                for metric in category:
+                    fd.write("{}\n".format(metric))
+
+    translate.translate_metrics(
```
Dexterp37: Not for this PR, if you don't have time, but it would be very nice to have a simple example like this one in the docs :-)

PR #179, opened 2020-02-28
MAINT: Add mypy type annotations.

mdboom:
Additionally, this fixes the black formatter check in CI and fixes a few
black violations that have slipped through.

### DIFF
--- a/glean_parser/kotlin.py
+++ b/glean_parser/kotlin.py
@@ -11,11 +11,15 @@ Outputter to generate Kotlin code for metrics.
 from collections import OrderedDict
 import enum
 import json
+from pathlib import Path
+from typing import Any, Dict, List, Union  # noqa
 
+from . import metrics
+from . import pings
 from . import util
 
 
-def kotlin_datatypes_filter(value):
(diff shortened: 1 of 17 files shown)


glean_parser/kotlin.py
```diff
@@ -11,11 +11,15 @@
...
+from pathlib import Path
+from typing import Any, Dict, List, Union  # noqa
 
+from . import metrics
+from . import pings
 from . import util
 
 
-def kotlin_datatypes_filter(value):
+def kotlin_datatypes_filter(value: util.JSONType) -> str:
```
Dexterp37: Oh, type annotations are ok in Python 3.5, but not variable annotations? Interesting

mdboom: Yeah -- the syntax for annotations on function arguments was added in 3.5, but annotation local variables was added in 3.6.  I think it was originally assumed that you would only want to type check on function boundaries, but it turns out that's not enough.  (The whole thing is awful in Python, honestly, but I do think it provides *some* value -- it detected a few type-correctness bugs found and fixed in this PR, for example).

PR #183, opened 2020-04-14
Bug 1589635: Treat linter warnings as errors in translate command.

mdboom:
Also breaks out the concept of glinter warnings and errors so we can
add checks that don't break the build in the future.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,7 +5,8 @@ History
 Unreleased
 ----------
 
-* `glean_parser` now produces a linter error when `user` lifetime metrics are
+* **Breaking change:** glinter errors found during the `translate` command will now return an error code. glinter warnings will be displayed, but not return an error code.
+* `glean_parser` now produces a linter warning when `user` lifetime metrics are
   set to expire. See [bug 1604854](https://bugzilla.mozilla.org/show_bug.cgi?id=1604854)
   for additional context.
 
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
+import enum
 from pathlib import Path
 import re
 import sys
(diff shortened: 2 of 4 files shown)


HISTORY.rst
```diff
@@ -5,6 +5,8 @@ History
 Unreleased
 ----------
 
+* **Breaking change:** glinter nits found during the `translate` command are now treated as errors, not warnings.
```
Dexterp37, line 8: Let's also call out that glinter `errors` are treated as errors, not `warnings` (even though we don't have them at the moment). Are you planning on having a follow-up bug to discuss what's an error and what's a warning?

mdboom: My thinking is that we would roll out new nits as warnings, and gradually move them to errors after users have had a chance to catch up.  It's possible on a case-by-case basis that some things would remain errors forever.

glean_parser/lint.py
```diff
@@ -194,31 +194,50 @@ def check_misspelled_pings(
                 )
 
 
+# The checks that operate on an entire category of metrics:
+#    {NAME: (function, is_error)}
 CATEGORY_CHECKS = {
-    "COMMON_PREFIX": check_common_prefix,
-    "CATEGORY_GENERIC": check_category_generic,
-}  # type: Dict[str, Callable[[str, Iterable[metrics.Metric]], LintGenerator]]
+    "COMMON_PREFIX": (check_common_prefix, True),
```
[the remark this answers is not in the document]
mdboom: That's a good idea.

HISTORY.rst
Dexterp37: That's fine! But in the long run, we're not going to break translate if we add new "warning classes", right?

mdboom: That's right.  Warnings won't cause translate to return an error code from the process.

Dexterp37: Cool, I'd just explicitly mention that in the changelog, for clarity for people outside of our group.

mdboom:
I rebased this to include the new user lifetime expiration check, and made it a warning, so we can roll this out without breaking the build for folks.

PR #184, opened 2020-04-20
Bug 1631078: Loosen minimum versions of dependencies and test them in CI

mdboom:
This lowers the minimum versions of dependencies as far as we can without breaking our unit tests.  Then it adds this to CI to make sure we don't start using newer APIs accidentally without updating the minimum versions.  (The third-party "requirements-builder" tool is used to help with this process).

This is a prerequisite to doing the same thing in Glean SDK.

This shouldn't have any bearing on the vendoring in m-c.

### DIFF
--- a/glean_parser/util.py
+++ b/glean_parser/util.py
@@ -253,6 +253,10 @@ def fetch_remote_url(url: str, cache: bool = True):
     Fetches the contents from an HTTP url or local file path, and optionally
     caches it to disk.
     """
+    # Include the Python version in the cache key, since caches aren't
+    # sharable across Python versions.
+    key = (url, str(sys.version_info))
+
     is_http = url.startswith("http")
 
     if not is_http:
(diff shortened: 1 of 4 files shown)


glean_parser/util.py
```diff
@@ -253,6 +253,10 @@ def fetch_remote_url(url: str, cache: bool = True):
     Fetches the contents from an HTTP url or local file path, and optionally
     caches it to disk.
     """
+    # Include the Python version in the cache key, since caches aren't
+    # sharable across Python versions.
+    key = (url, str(sys.version_info))
```
mdboom, line 258: This isn't strictly related, but I did discover it while testing on a bunch of different Python versions locally (since they share the same cache files).

mdboom:
> +1 I did not know about requirements builder!

Neither did I until I went looking for it :)

PR #189, opened 2020-05-12
Bug 1635754 - Explicitly test that underscores in ping names are disallowed

badboy:

### DIFF
--- a/tests/test_swift.py
+++ b/tests/test_swift.py
@@ -53,7 +53,7 @@ def test_parser(tmpdir):
     # Only run this test if swiftc is on the path.
     if shutil.which("swiftc"):
         for filepath in tmpdir.glob("*.swift"):
-            subprocess.check_call(["swiftc", "-dump-parse", filepath])
+            subprocess.check_call(["swiftc", "-parse", filepath])
 
     # Lint check on the generated files.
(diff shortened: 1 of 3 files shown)


tests/test_swift.py
```diff
@@ -53,7 +53,7 @@ def test_parser(tmpdir):
     # Only run this test if swiftc is on the path.
     if shutil.which("swiftc"):
         for filepath in tmpdir.glob("*.swift"):
-            subprocess.check_call(["swiftc", "-dump-parse", filepath])
+            subprocess.check_call(["swiftc", "-parse", filepath])
```
mdboom, line 56: What's the context on this change?  Also, I don't think this runs in CI (which is currently Linux-only), and maybe we should address that at some point.

badboy: I _think_ that locally my Swift updated and thus causing test failures when it uses `-dump-parse` (as it expects `GleanMetrics` and other symbols to be present). `-parse` on the other hand really just parses, without looking up symbols.

We once did think about getting that test on CI, but

* getting swift on Linux is ... not easy
* running on macOS is more expensive

badboy:
I guess it'd be good to actually include the file, right?

PR #191, opened 2020-05-19
Swift: Combine all metrics and pings into a single generated file

badboy:
I tested this locally against Glean and the sample app.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,7 +8,8 @@ Unreleased
 * `glinter` messages have been improved with more details and to be more
   actionable.
 * A maximum of 10 `extra_keys` is now enforced for `event` metric types.
-* BUGFIX: the `Lifetime` enum values now match the values of the implementation in mozilla/glean. 
+* BUGFIX: the `Lifetime` enum values now match the values of the implementation in mozilla/glean.
+* **Breaking change:** (Swift only) Combine all metrics and pings into a single generated file `Metrics.swift`.
 
 1.20.4 (2020-05-07)
 -------------------
--- a/glean_parser/swift.py
+++ b/glean_parser/swift.py
@@ -16,7 +16,6 @@ from typing import Any, Dict, Union
 from . import metrics
 from . import pings
 from . import util
-from collections import defaultdict
 
 # An (imcomplete) list of reserved keywords in Swift.
 # These will be replaced in generated code by their escaped form.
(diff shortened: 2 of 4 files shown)


glean_parser/swift.py
```diff
@@ -107,6 +107,18 @@ def variable_name(var: str) -> str:
...
+class Metric:
+    """
+    Data struct holding information about a metric to be used in the template.
+    """
+
+    category_name = None  # type: str
+    objs = None  # type: Dict[str, Union[metrics.Metric, pings.Ping]]
+    is_ping_type = False  # type: bool
+
+    pass
```
mdboom, line 119: `pass` isn't necessary, since there are some members defined.

glean_parser/swift.py
```diff
@@ -107,6 +107,18 @@ def variable_name(var: str) -> str:
         return var
 
 
+class Metric:
```
mdboom, line 110: Doesn't this contain a "category" of metrics, not a single metric?  Maybe name `Category` or `MetricCollection`?

glean_parser/swift.py
```diff
@@ -149,31 +161,32 @@ def output_swift(
...
-        filepath = output_dir / filename
+    filename = "Metrics.swift"
+    filepath = output_dir / filename
+    metrics = []
 
+    for category_key, category_val in objs.items():
         custom_pings = defaultdict()  # type: Dict[str, pings.Ping]
         for obj in category_val.values():
             if isinstance(obj, pings.Ping):
                 custom_pings[obj.name] = obj
```
mdboom, line 172: Maybe just `contains_pings = any(isinstance(obj, pings.Ping) for obj in category_val.values())`?

glean_parser/swift.py
```diff
@@ -107,6 +107,18 @@ def variable_name(var: str) -> str:
...
 
 
+class Metric:
+    """
+    Data struct holding information about a metric to be used in the template.
+    """
+
+    category_name = None  # type: str
+    objs = None  # type: Dict[str, Union[metrics.Metric, pings.Ping]]
+    is_ping_type = False  # type: bool
```
mdboom, line 117: `in_ping_type` -> `contains_pings`?

HISTORY.rst
```diff
@@ -7,6 +7,7 @@ Unreleased
 
 * `glinter` messages have been improved with more details and to be more
   actionable.
+* Swift: Combine all metrics and pings into a single generated file `Metrics.swift`.
```
mdboom, line 10: Should this be flagged as a breaking change, since it changes who users need to import their metrics?  (If I'm right about that...)

glean_parser/swift.py
badboy: Ah, before I didn't have members so I needed it.

glean_parser/swift.py
badboy: Hm, yeah, actually the naming in below code confused me enough to not be able to come up with a good name here.
I'll swap it to one of your suggestions.

glean_parser/swift.py
badboy: Hm, makes sense.

HISTORY.rst
badboy: Ah yes, true indeed.
Once we upgrade the parser in the SDK we should make sure to change the docs

PR #198, opened 2020-07-13
Add support for generating C# code

Dexterp37:
This adds support for all the metric types, but it may fail for events, in the future.
That's fine, as we don't currently support events in C# and we'll take care of that in Bug 1648422.


glean_parser/csharp.py
```diff
@@ -0,0 +1,166 @@
...
+        "csharp.jinja2",
+        filters=(
+            ("csharp", csharp_datatypes_filter),
+            ("type_name", type_name),
+            ("class_name", class_name),
+        ),
+    )
+
+    # The object parameters to pass to constructors
+    extra_args = [
```
mdboom, line 130: I'm not a fan of this duplication, but we should tackle that globally: https://bugzilla.mozilla.org/show_bug.cgi?id=1652471

glean_parser/csharp.py
```diff
@@ -0,0 +1,166 @@
...
+      - lists to use `new string[] {}` (only strings)
+      - dicts to use mapOf (not currently supported)
+      - sets to use setOf (not currently supported)
+      - enums to use the like-named C# enum
+    """
+
+    class CSharpEncoder(json.JSONEncoder):
+        def iterencode(self, value):
+            if isinstance(value, list):
+                yield "new string[] {"
```
mdboom, line 35: Since this will break if passed a list not containing all strings, maybe we add an assertion here?

```
assert all(isinstance(x) for x in value)
```
?

glean_parser/templates/csharp.jinja2
```diff
@@ -0,0 +1,97 @@
...
+ * AUTOGENERATED BY glean_parser.  DO NOT EDIT.
+ */
+{# The rendered markdown is autogenerated, but this
+Jinja2 template is not. Please file bugs! #}
+
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+{% macro obj_declaration(obj, suffix='', access='', lazy=False) %}
+{{ access }} {% if lazy %} Lazy<{{ obj|type_name }}>{%- else %} {{ obj|type_name }}{% endif %} {{ obj.name|camelize }}{{ suffix }}{% if lazy %} = new Lazy<{{ obj|type_name }}>(() => {%- else %} ={% endif %}
```
mdboom, line 13: nit (only if not a quagmire): You can use `{%-` and `-%}` around commands to remove whitespace which means you don't have to have a big run-on line like this.

tests/test_csharp.py
```diff
@@ -0,0 +1,203 @@
...
+from glean_parser import pings
+from glean_parser import translate
+
+
+ROOT = Path(__file__).parent
+
+
+# The version of Detekt built. Must match the version checked out in
+# Makefile
+DETEKT_VERSION = "1.1.1"
```
mdboom, line 20: Not necessary for C#

glean_parser/csharp.py
Dexterp37: Oh, thanks for doing this!

PR #199, opened 2020-07-16
Bug 1608838: Include data collection categories as metadata

mdboom:
This work may be necessary to feed into data access policies in the pipeline. Also, it would be a cool thing to start including in the ongoing Glean Dictionary work.

The included script works great against Fenix' metrics (though it did find one case where a data review is missing), so the plan would be to update them in an automated fashion once we work out the design details here.


tools/extract_data_categories.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env python3
```
Dexterp37, line 1: nit: this is missing the license header

glean_parser/markdown.py
```diff
@@ -129,6 +129,13 @@ def ping_include_client_id(
         return False
 
 
+def data_category_numbers(data_category: Optional[List[metrics.DataCategory]]) -> str:
```
Dexterp37, line 132: @mdboom can you kindly add an HISTORY entry?

mdboom:
> r+ with the nit addressed. Is there a long-term plan or strategy for this field? Do we plan on eventually making it mandatory ? (may be very Mozilla specific, though :) )

The plan is to learn from having this field and expose it in the Glean Dictionary, and also experiment with using it to set permissions on tables (or at least to catch cases where we may have made a mistake).

I think making it mandatory is TBD -- we would at least need to update all of our users' products before we could make that happen.  But it's kind of cool how easy it was to do that be scraping existing reviews...

mdboom:
Comment from @hamilton: Maybe `data_sensitivity` rather than `data_category`?

jklukas:
This work definitely has some parallels and potential connections to the current [proposal for encoding pipeline metadata in JSON schemas](https://docs.google.com/document/d/1oNi-eX_IPIZa8C0h5GZnWTM0WFivZ1hPMgMMX3c3HHs/edit?ts=5f10bdeb#) although that document is focusing on doctype-level or table-level metadata for now, with field-level metadata as a future concern.

I could definitely imagine having the schema generator ingest these `data_category` fields and inject them into the generated schemas for Glean doctypes, which would in turn make the information available to the pipeline and to ops logic.

Naming things is always hard, but I do like the use of an enum here with logical names for the categories rather than relying on integers.

We have not yet investigated thoroughly, but BigQuery does now support [column-level security restrictions](https://cloud.google.com/bigquery/docs/column-level-security) as a beta feature. It interacts with Data Catalog, and expects you to define your own security taxonomy there which can then be applied to fields. So I could certainly imagine a future where the `data_category` values defined here are passed to Data Catalog in ops logic, so that the category defined on a field has a concrete effect on what audiences are able to view the field in BigQuery. We have no concrete plans to do so at this point, though.

PR #201, opened 2020-07-20
Bug 1650787 - Support JWE metric types

brizental:
* Create JweMetric objects from Swift, Kotlin and C#;
* Display JWE metrics in markdown with `decrypted_name` as an extra.

### DIFF
--- a/glean_parser/schemas/metrics.1-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.1-0-0.schema.yaml
@@ -97,6 +97,8 @@ definitions:
 
             - `uuid`: Record a UUID v4.
 
+            - `jwe`: Record a JWE value.
+
             - `memory_distribution`: A histogram for recording memory usage
               values. Additional properties: `memory_unit`.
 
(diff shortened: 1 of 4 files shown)


glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -412,6 +415,16 @@ definitions:
...
 
+      decrypted_name:
+        title: Decrypted name
+        description: |
+          Name of the column where to persist the decrypted value
+          stored in the JWE after processing.
+
+          Required when `type`_ is `jwe`.
+        type: string
+        pattern: "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})*$"
```
brizental, line 426: This is the same regex from `dotted_snake_case`.

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -97,6 +97,8 @@ definitions:
 
             - `uuid`: Record a UUID v4.
 
+            - `jwe`: Record a JWE value.
```
mdboom, line 100: Nit: Maybe link to the JWE standard here?  (We can use markdown in this file...)

PR #203, opened 2020-07-22
Bug 1654645 - fix `deletion_request` to `deletion-request` in reserved ping names

travis79:


mdboom:
To confirm: No issues with probe-scraper.  The metric is correctly merged together with both the `-` and `_` naming of the ping.

PR #222, opened 2020-08-26
Use bugbear in CI

mdboom:


badboy (changes requested):
I assume because flake8-bugbear is a plugin it just runs automatically as part of the flake8 invocation?

And I also assume that all code changes are applying the suggested changes of that plugin?

Let's use the commit message to state those things!

badboy (approved):

PR #224, opened 2020-09-01
Bug 1599764 -  Have glinter warn when expiry date is too far in the future

brizental:

### DIFF
--- a/glean_parser/util.py
+++ b/glean_parser/util.py
@@ -336,6 +336,25 @@ def format_error(filepath: Union[str, Path], header: str, content: str) -> str:
         return f"{filepath}:\n{_utils.indent(content)}"
 
 
+def parse_expires(expires: str) -> datetime.date:
+    """
+    Parses the expired field date (yyyy-mm-dd) as a date.
+    Raises a ValueError in case the string is not properly formatted.
+    """
+    try:
+        if sys.version_info < (3, 7):
+            return iso8601.parse_date(expires).date()
+        else:
+            return datetime.date.fromisoformat(expires)
+    except ValueError:
+        # The schema enforces this formatting for the expires field,
+        # we shouldn't ever get to this exception.
+        raise ValueError(
+            f"Invalid expiration date '{expires}'. "
+            "Must be of the form yyyy-mm-dd in UTC."
+        )
+
+
 def is_expired(expires: str) -> bool:
     """
     Parses the `expires` field in a metric or ping and returns whether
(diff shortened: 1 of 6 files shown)


glean_parser/util.py
```diff
@@ -346,29 +365,24 @@ def is_expired(expires: str) -> bool:
...
-                "Must be of the form yyyy-mm-dd in UTC."
-            )
+        date = parse_expires(expires)
         return date <= datetime.datetime.utcnow().date()
 
 
 def validate_expires(expires: str) -> None:
     """
-    Raises ValueError if `expires` is not valid.
+    Raises a ValueError in case the date is more than 730 days (~2 years) in the future
```
chutten, line 374: or if `expires` isn't ISO8601 parseable.

glean_parser/util.py
```diff
@@ -346,29 +365,24 @@ def is_expired(expires: str) -> bool:
...
-        iso8601.parse_date(expires)
-    else:
-        datetime.date.fromisoformat(expires)
+
+    date = parse_expires(expires)
+    max_date = datetime.datetime.now() + datetime.timedelta(days=730)
+    if date > max_date.date():
+        raise ValueError(
+            f"Invalid expiration date '{expires}'. '"
+            "Date must not be more than 730 days (~2 years) in the future."
```
chutten, line 384: I think we should be a little kinder in the language here. A date more than 2 years in the future isn't invalid as much as it is a likely error.

Perhaps `"Expiration date '{expires}' is more than 730 days (2 years) in the future. Please make sure this is intentional."`? What do you think?

tests/test_lint.py
```diff
@@ -266,3 +267,29 @@ def test_expired_metric():
 
     assert len(nits) == 1
     assert set(["EXPIRED"]) == set(v.check_name for v in nits)
+
+
+def test_expires_too_far_in_the_future():
+    """Test that a misformatted `expires` field generates errors"""
```
chutten, line 273: It isn't misformatted as much as it is just very far in the future.

Also maybe put in a note that this test is expected to start failing in 2098 : D

glean_parser/util.py
brizental: Yeap, I had that there, but I removed it because we should never get this errors since it is validated by the schema. Since we _could_ throw, according tot he code, I'll add it back.

tests/test_lint.py
brizental: Yup, for got to change this docstring, oops.

> Also maybe put in a note that this test is expected to start failing in 2098 : D

LOL, sure.

glean_parser/util.py
mdboom: FWIW, there is a message and the bottom of `glinter` output about the `no_lint` option that's always displayed when there are 1 or more warnings or errors.  So *maybe* no need to explicitly have that here, unless we **really** want to highlight that.

brizental: Uh oh, this was alredy merged with the message and I was not aware of that. I can quickly remove the message if you think its best. https://github.com/mozilla/glean_parser/blob/main/glean_parser/util.py#L386-L387

mdboom: Not critical -- I didn't read the final diff when I wrote the comment.  We can remove later if it annoys anyone (but it probably won't).

PR #225, opened 2020-09-02
Bug 1658349 - Allow using the quantity metric type outside of Gecko

brizental:

### DIFF
--- a/glean_parser/schemas/metrics.1-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.1-0-0.schema.yaml
@@ -83,8 +83,7 @@ definitions:
 
             - `counter`: A numeric value that can only be incremented.
 
-            - `quantity`: A numeric value that is set directly. Only allowed for
-              metrics coming from GeckoView.
+            - `quantity`: A numeric value that is set directly.
 
             - `timespan`: Represents a time interval. Additional properties:
               `time_unit`.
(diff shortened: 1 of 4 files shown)


glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -548,12 +547,11 @@ additionalProperties:
             type:
               enum:
                 - custom_distribution
-                - quantity
         then:
           required:
             - gecko_datapoint
           description: |
-            `custom_distribution` and `quantity` is only allowed for Gecko
+            `custom_distribution` is only allowed for Gecko
```
Dexterp37, line 554: Did you check if you're allowed to use `quantity` without a `gecko_datapoint` property?

brizental: Manually yes, do you think I should add a unit test?

Dexterp37: Nope, that's fine. If you checked manually and it works, I trust you :)

Dexterp37 (approved):

glean_parser/schemas/metrics.1-0-0.schema.yaml
brizental: Also the `test_quantity` test kinda implicity tests this since it now only gets one error and not two. Which reminds I should update it's docstring.

PR #227, opened 2020-09-11
Bug 1664306 - Add Custom Expiry support

chutten:

### DIFF
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -46,7 +46,7 @@ class Metric:
         bugs: List[str],
         description: str,
         notification_emails: List[str],
-        expires: str,
+        expires: Any,
         data_reviews: Optional[List[str]] = None,
         version: int = 0,
         disabled: bool = False,
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -56,7 +56,7 @@ class Metric:
         gecko_datapoint: str = "",
         no_lint: Optional[List[str]] = None,
         data_sensitivity: Optional[List[str]] = None,
-        _config: Optional[Dict[str, Any]] = None,
+        _config: Dict[str, Any] = None,
         _validated: bool = False,
     ):
         # Avoid cyclical import
@@ -97,6 +97,11 @@ class Metric:
             }  # type: Dict[str, util.JSONType]
             for error in parser.validate(data):
                 raise ValueError(error)
+
+        # Store the config, but only after validation.
+        if _config is None:
+            _config = {}
+        self._config = _config
 
         # Metrics in the special category "glean.internal.metrics" need to have
         # an empty category string when identifying the metrics in the ping.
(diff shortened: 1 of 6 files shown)


glean_parser/metrics.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         gecko_datapoint: str = "",
         no_lint: Optional[List[str]] = None,
         data_sensitivity: Optional[List[str]] = None,
-        _config: Optional[Dict[str, Any]] = None,
+        _config: Dict[str, Any] = {},
```
mdboom, line 59: Python nonsense: Shouldn't use a mutable data structure as a default keyword argument because it will use the same object for every invocation.

The usual workaround pattern is to set `None` as the default and then have:

```python
if _config is None:
    _config = {}
```

in the body.

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -239,16 +239,6 @@ definitions:
         description: |
           **Required.**
 
-          May be one of the following values:
```
mdboom, line 242: I still think we should document the default behavior here.  Maybe retain this, but add a note that the default behavior might be overridden in certain applications ... see the `custom_validate_expires` config.

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -239,16 +239,6 @@ definitions:
...
-            - `<build date>`: An ISO date `yyyy-mm-dd` in UTC on which the
-              metric expires. For example, `2019-03-13`. This date is checked at
-              build time. Except in special cases, this form should be used so
-              that the metric automatically "sunsets" after a period of time.
-            - `never`: This metric never expires.
-            - `expired`: This metric is manually expired.
-        type: string
```
mdboom, line 249: I think we should still enforce `type: string` if we can.  Accepting anything could let through some hard-to-diagnose YAML silliness where it thinks things aren't strings.  If build id really will be a number, we could do:

```
anyOf:
    - type: string
    - type: number
```

(but I would argue build ids aren't numbers, even if they are composed of only numeric characters, and systems that assume they are always numbers lead to [this kind of thing](https://github.com/mozilla-mobile/fenix/blob/e6ee13dc/buildSrc/src/main/java/Config.kt#L55-L112).

glean_parser/metrics.py
mdboom: (Aside -- the `bugbear` linter slated to turn on in #222 would have caught this sort of thing...)

mdboom (changes requested):
Welcome to `glean_parser`, @chutten!

glean_parser/schemas/metrics.1-0-0.schema.yaml
chutten: Fair enough. Strings it is.

glean_parser/schemas/metrics.1-0-0.schema.yaml
chutten: Good idea.

chutten:
RFAL, @mdboom

mdboom (approved):
LGTM

PR #231, opened 2020-09-14
Bug 1664889: Display glean_parser version when there is an error

mdboom:

### DIFF
--- a/glean_parser/__main__.py
+++ b/glean_parser/__main__.py
@@ -131,5 +131,18 @@ main.add_command(check)
 main.add_command(glinter)
 
 
+def main_wrapper(args=None):
+    """
+    A simple wrapper around click's `main` to display the glean_parser version
+    when there is an error.
+    """
+    try:
+        main(args=args)
+    except SystemExit as e:
+        if e.code != 0:
+            print(f"ERROR running glean_parser v{glean_parser.__version__}")
+        raise
+
+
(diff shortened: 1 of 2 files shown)


glean_parser/__main__.py
```diff
@@ -131,5 +131,18 @@ def main(args=None):
...
+        main(args=args)
+    except SystemExit as e:
+        if e.code != 0:
+            print(f"ERROR running glean_parser v{glean_parser.__version__}")
+        raise
+
+
 if __name__ == "__main__":
-    sys.exit(main())  # pragma: no cover
+    main_wrapper()  # pragma: no cover
```
chutten, line 148: I know Python is magic and all, but `sys.exit` takes a number, but `main_wrapper` acts like `main` throws an exception and returns void. Which is it?

chutten: (Not discounting the possibility is that the answer is "Both."

chutten (approved):

glean_parser/__main__.py
mdboom: Thanks for calling this out.  I think actually click changed their API.  It seems to always throw now according to the docs, and the `sys.exit(main())` pattern is no longer recommended on the [front page](https://click.palletsprojects.com/en/7.x/).  So, good to go, I think.

PR #235, opened 2020-10-01
Bug 1668547: Incl. metric identifier in comment in gen. code

mdboom:

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,7 @@ Unreleased
 ----------
 
 * **Breaking change:** `glean_parser` will now return an error code when any of the input files do not exist (unless the `--allow-missing-files` flag is passed).
+* Generated code now includes a comment next to each metric containing the name of the metric in its original `snake_case` form.
 
 1.28.6 (2020-09-24)
 -------------------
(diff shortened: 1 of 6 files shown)


glean_parser/pings.py
```diff
@@ -73,3 +73,9 @@ def serialize(self) -> Dict[str, util.JSONType]:
         d = self.__dict__.copy()
         del d["name"]
         return d
+
+    def identifier(self) -> str:
```
chutten, line 77: pings.py? Shouldn't this be on Metric, not Ping?

chutten (commented):
Going to add it to C# as well?

glean_parser/pings.py
mdboom: The same template is used to generate ping objects and metric objects, so we need `identifier` defined on both.  (Alternatively, could have put this logic in the templates, but that was much more complex...)

mdboom:
> Going to add it to C# as well?

Yeah, I guess I might as well (even though that's on pause...)

HISTORY.rst
```diff
@@ -5,6 +5,8 @@ History
 Unreleased
 ----------
 
+* Generated code now includes a comment next to each metric containing the name of the metric in its original `snake_case` form.
```
chutten, line 8:
```suggestion
* Generated code now includes a comment next to each metric containing the name of the metric in its original `dotted.snake_case` form.
```

chutten (approved):

PR #251, opened 2020-12-01
Add line number to metrics and pings

Iinh:
Gather the line number in yaml files where metrics and pings are defined, as the first step of [glean-dictionary/issues/223](https://github.com/mozilla/glean-dictionary/issues/223).


glean_parser/util.py
```diff
@@ -66,13 +66,35 @@ def remove_implicit_resolver(cls, tag_to_remove):
...
+        line_number = {"line": str(node.start_mark.line)}
+
+        for value in node.value:
+            if value[0].value != "$schema":
+                # pings
+                if node.start_mark.column == 2 and value[0].value == "description":
+                    mapping['defined_in'] = line_number
+                # metrics
+                if node.start_mark.column == 4 and value[0].value == "type":
+                    mapping['defined_in'] = line_number
```
Iinh, line 87: The logic used here is admittedly hacky, but right now I can't think of a better way

mdboom: Yeah -- this stuff is always hard and hacky to do.  Rather than focusing on the specific fields `description` and `type` and column numbers, what if we just add the first line we see to every mapping, something like:

```
mapping['defined_in`] = min(line_number, mapping.get('defined_in', sys.maxint))
```

mdboom (commented):
This looks great, and will be really useful.  My comments are mainly just from you hitting uncharted territory here...

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -145,6 +145,11 @@ definitions:
           syntax](https://www.markdownguide.org/basic-syntax/).
         type: string
 
+      defined_in:
```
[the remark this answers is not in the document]
wlach: @mdboom That approach makes sense to me, I can't think of anything better offhand.

Iinh:
Hi @mdboom! Thank you for your review! I’ve updated the code with your suggestions, if this looks okay I’ll go ahead and add tests. Thanks!

glean_parser/util.py
Iinh: @mdboom The reason why the columns number, and value checking of “description” or “type” were used is because due to the differences in the yaml file structures for pings and metrics, there’s only a few very specific places where we can map `defined_in` to:

- pings.yaml: we want to map only when conditions are node `column == 2` and `value == “description”` to make sure that a) it is a ping we’re mapping to, and b) `defined_in` will be at the same level as the other ping properties. 

- metrics.yaml: we don’t want to map `defined_in` at `column == 2` because it will map it at the same node level as the other metrics, which `defined_in` then becomes another metric instead of just a metric property. `column == 4` and `value == “type”` will make sure that `defined_in` will be mapped at the metric property level. These 2 conditions also help to avoid mapping to nodes like [this one](https://github.com/mozilla/glean/blob/1da5cdf7300e814bd5993652b0a60a11c45c5ed6/glean-core/pings.yaml#L30) whose column also starts at 4.

I hope that made sense. I did try what you suggested, but I’ve found that without being very specific with the mapping condition it will cause errors.

glean_parser/metrics.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         description: str,
         notification_emails: List[str],
         expires: Any,
+        defined_in: Optional[str] = None,
```
mdboom, line 50: *Very small nit*: Can this go at the end of the argument list (after `data_sensitivity`, but before `_config`) so this "output only" thing is separate from the ones we expect to find in the file?

glean_parser/parser.py
```diff
@@ -88,7 +89,8 @@ def _load_file(
 
     filetype = FILE_TYPES.get(schema_key)
 
-    for error in validate(content, filepath):
+    modified_content = util.remove_output_params(copy.deepcopy(content), "defined_in")
```
mdboom, line 92: Why does this need to be removed here?  This file *should* only affect things coming directly from an input file.  (I'm probably just not being creative enough about where this is called from: can you share where this breaks if you *don't* make this change?)

glean_parser/pings.py
```diff
@@ -24,6 +24,7 @@ def __init__(
         description: str,
         bugs: List[str],
         notification_emails: List[str],
+        defined_in: Optional[str] = None,
```
mdboom, line 27: Same small nit as above: Can we move this near the end?

glean_parser/schemas/metrics.1-0-0.schema.yaml
```diff
@@ -144,7 +144,7 @@ definitions:
           Descriptions may contain [markdown
           syntax](https://www.markdownguide.org/basic-syntax/).
         type: string
-
+
```
mdboom, line 147: Über-nit: Let's just remove this whitespace-only change altogether.

glean_parser/util.py
```diff
@@ -399,6 +420,19 @@ def report_validation_errors(all_objects):
     return found_error
 
 
+def remove_output_params(d, output_params):
+    """
+    Removes output-only params, such as "defined_in",
+    in order to validate the output against the input schema.
+    """
+    for key, value in copy.deepcopy(d).items():
```
mdboom, line 428: This is actually editing the passed-in dictionary in-place, which could be a problem if the caller is depending on the dictionary not changing.  Also, I don't think we need to recurse down -- we only need to remove certain keys at the top-level.  Therefore, we could just use a dictionary comprehension (which will make a new dictionary, rather than modifying in-place):

```python
return { k, v for k, v in d.items() if k not in output_params }
```

glean_parser/util.py
```diff
@@ -66,13 +67,33 @@ def remove_implicit_resolver(cls, tag_to_remove):
...
+        mapping = super(SafeLineLoader, self).construct_mapping(
+            node, deep=deep)
+        for value in node.value:
+            if value[0].value != "$schema":
+                # pings
+                if node.start_mark.column == 2 and value[0].value == "description":
+                    mapping['defined_in'] = str(node.start_mark.line)
+                # metrics
+                if node.start_mark.column == 4 and value[0].value == "type":
+                    mapping['defined_in'] = str(node.start_mark.line)
```
mdboom, line 86: I think I understand why this is so hard now, thanks to your response ;)  It seems a bit brittle (as in, what if we change the names of the fields in future), but hopefully some good unit tests will help us catch that.

Can you maybe add a comment to this function about why this approach is used (kind of like your comment to my question in the PR)?  Also, let's note that the column numbers work because we run `yamllint` on the input that enforces 2-space tab width -- without that, YAML writers could use whatever tab width they want and this would break.

mdboom (changes requested):
This is great work!  Just a few more comments.

glean_parser/util.py
```diff
@@ -120,15 +150,20 @@ def load_yaml_or_json(path: Path, ordered_dict: bool = False):
...
             if ordered_dict:
-                return ordered_yaml_load(fd)
+                content = ordered_yaml_load(fd)
             else:
-                return yaml.load(fd, Loader=_NoDatesSafeLoader)
+                content = yaml.load(fd, Loader=SafeLineLoader)
+        if add_line_number:
+            return content
+        else:
+            return remove_output_params(content, "defined_in")
```
Iinh, line 166: `add_line_number` param was added here to clarify when [we need to add line number to the output](https://github.com/Iinh/glean_parser/blob/0e95f993a4a715c467c405f6b673b04532058b51/glean_parser/parser.py#L64-L67) versus [when we don't](https://github.com/Iinh/glean_parser/blob/0e95f993a4a715c467c405f6b673b04532058b51/glean_parser/parser.py#L109).

glean_parser/util.py
Iinh: @mdboom We used this util function at various places: 
- if it's to [serialize pings and metrics]( https://github.com/Iinh/glean_parser/blob/0e95f993a4a715c467c405f6b673b04532058b51/glean_parser/metrics.py#L175) then yes we don’t need to recurse down
- however, we also used this function in [`_load_file`](https://github.com/Iinh/glean_parser/blob/0e95f993a4a715c467c405f6b673b04532058b51/glean_parser/parser.py#L93) and in [`load_yaml_or_json`](https://github.com/Iinh/glean_parser/blob/75dfc5e51925adf3bee4bdf763c626b29cb4024c/glean_parser/util.py#L149-L152) to do a bulk removal of all the line numbers because the content passed in is a multi-layered nested dict, so recursion is necessary

I've updated the function to not edit the passed-in dict while still keeping recursion per my explanation above. Looking forward to feedback!

glean_parser/parser.py
Iinh: @mdboom The content to be validated was loaded from ping and metric yaml files through [`load_yaml_or_json`](https://github.com/Iinh/glean_parser/blob/75dfc5e51925adf3bee4bdf763c626b29cb4024c/glean_parser/util.py#L149-L152), which uses `SafeLineLoader` that adds line number to the output. So to be able to successfully validate the content against the schema here, we need to remove the line number from the output first.

glean_parser/util.py
```diff
@@ -66,13 +66,40 @@ def remove_implicit_resolver(cls, tag_to_remove):
...
+
+
 if sys.version_info < (3, 7):
     # In Python prior to 3.7, dictionary order is not preserved. However, we
     # want the metrics to appear in the output in the same order as they are in
     # the metrics.yaml file, so on earlier versions of Python we must use an
     # OrderedDict object.
     def ordered_yaml_load(stream):
-        class OrderedLoader(_NoDatesSafeLoader):
+        class OrderedLoader(SafeLineLoader):
```
Iinh, line 102: This works fine for other Python versions, but for some reason OrderedLoader is not instantiated and used correctly in v3.6 which might have caused the failed build-36 test. I was wondering what you think is the best way to fix this. Thanks @mdboom

glean_parser/util.py
```diff
@@ -66,13 +66,40 @@ def remove_implicit_resolver(cls, tag_to_remove):
...
+
+        for value in node.value:
+            if value[0].value != "$schema":
+                # pings.yaml
+                if node.start_mark.column == 2 and value[0].value == "description":
+                    mapping["defined_in"] = {"line": str(node.start_mark.line)}
+                # metrics.yaml
+                if node.start_mark.column == 4 and value[0].value == "type":
+                    mapping["defined_in"] = {"line": str(node.start_mark.line)}
+        return mapping
```
mdboom, line 93: Thanks for your patience with this PR -- I think it's turned out to be more complicated than anyone expected.

I think the need to recursively delete all keys called `defined_in` is a problem -- what if a metric or ping name is called `defined_in`?  It would get silently deleted.

I think the root of the issue is that this is attaching data to the dictionary, but it's really something different -- it doesn't represent anything "literally" coming from the file.  A bunch of the other weirdness flows out from this, too.

I think, instead, what we should do is use a subclass of `dict` so we can attach this line number metadata without it being part of the dictionary itself.

Using this instead seems to work:

```
class SafeLineLoader(_NoDatesSafeLoader):
    pass


def _construct_mapping_adding_line(loader, node):
    loader.flatten_mapping(node)
    mapping = DictWithMetadata(loader.construct_pairs(node))
    mapping.defined_in = {"line": node.start_mark.line}
    return mapping


SafeLineLoader.add_constructor(
    yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, _construct_mapping_adding_line
)
```

This does mean that we'll also have to change the place where metrics are built slightly as well (in `make_metric`), so that we pull this data back out.  At the end of `make_metric`:

```
        return cls.metric_types[metric_type](
            category=category,
            name=name,
            defined_in=getattr(metric_info, "defined_in", None),
            _validated=validated,
            _config=config,
            **metric_info,
        )
```

Then, we shouldn't need to do anything special to validate the YAML loaded in (the JSON schema validator will safely ignore the extra information, because it's an *attribute* on the dictionary, not a *key*).

We'll still need to do the thing where we delete the `defined_in` key in `_serialize_input`, but we don't need to do the "delete it at any level" thing, so it should be safe.

I think this also addresses the hard coding of "description" and "type" as special keys to identify where metrics/ping appear in the file, too.

Sorry to come up with a completely different solution so late in the game -- sometimes these things don't become clear until they are worked all the way through, so no wasted effort here at all.

tests/test_translate.py
```diff
@@ -182,3 +182,17 @@ def external_translator(all_objects, output_dir, options):
...
+
+def test_loading_line_number():
+    pings = load_yaml_or_json(ROOT / "data" / "pings.yaml", add_line_number=True)
+    metrics = load_yaml_or_json(ROOT / "data" / "core.yaml", add_line_number=True)
+    invalid = load_yaml_or_json(
+        ROOT / "data" / "invalid_pings.yaml", add_line_number=True
+    )
+    no_line_number_pings = load_yaml_or_json(ROOT / "data" / "pings.yaml")
+
+    assert pings["custom-ping"]["defined_in"]["line"] == "7"
```
mdboom, line 195: Is there a reason why the line numbers are strings and not numbers here?  Without a good reason, I think they'd be more natural as numbers (i.e. change `"7"` to `7` here and remove the place where the line numbers are converted to `str` above).

mdboom (changes requested):
Long comment inline.

I think the issue with Python 3.6 might be:

On Python 3.6, dictionaries don't preserve the order of the elements as they are added.  So we have a special hack to use `OrderedDict` rather than just `dict`.  The line number handling work here happens *before* that, and it uses regular `dict`s, so it loses the ordering before it has a chance to get to the workaround that preserves the ordering.  I think changing the ordering of inheritance of the various loader classes should fix that.  Additionally, in my suggestion for a new way of handling the line numbers, we'll probably need to subclass `OrderedDict` rather than `dict` when on Python 3.6.

Thanks for your patience -- This is good work!

glean_parser/parser.py
```diff
@@ -62,7 +62,7 @@ def _load_file(
     `parser_config["allow_missing_files"]` is `True`.
     """
     try:
-        content = util.load_yaml_or_json(filepath, ordered_dict=True)
+        content = util.load_yaml_or_json(filepath)
```
Iinh, line 65: `ordered_dict` was removed because with the new yaml loader, OrderedDict is always used as the new dict subclass

tests/test_parser.py
```diff
@@ -134,38 +119,144 @@ def test_parser_schema_violation():
         'very_long_metric_name_this_is_too_long_s_well' is too long
         """,
     ]
-
     # The validator reports a different error based on the python version, so
     # we need to provide two copies for this to work.
     if sys.version_info < (3, 7):
-        expected_errors.append(
-            """
+        expected_errors.extend(
```
Iinh, line 125: Now that we use OrderedDict for all Python versions, the error message changed slightly too

tests/test_translate.py
Iinh: This is just a habit from my old Excel days, we always defaulted to using string if we don't plan on doing any calculations with the numbers, mostly to avoid formatting error. So thank you for pointing this out, I have updated it!

Iinh:
@mdboom I have updated the PR with your suggestions, please let me know if there's anything else that I might have missed. Michael, thank you so much for your help on this, especially your detailed explanation on why each change was necessary. I'm still new to Python, and working on this PR has definitely taught me a lot!

tests/test_parser.py
```diff
@@ -134,38 +119,144 @@ def test_parser_schema_violation():
...
-                  key_8:
-                    description: Sample extra key
-                  key_9:
-                    description: Sample extra key
-                  key_10:
-                    description: Sample extra key
-                  key_11:
-                    description: Sample extra key
+                event_too_many_extras:
+            extra_keys: !!python/object/apply:collections.OrderedDict
```
mdboom, line 215: This funny output (with the `python/object/apply:collections.OrderedDict` thing) is undesirable and confusing -- we want it to look the same as regular YAML input.  You may have to either -- go back to doing something different on Python 3.6, or use the `OrderedDumper` hack currently used only be Python 3.6 on all Python versions.

mdboom (changes requested):
This is way better, thanks!

Just one more detail around the weird looking YAML output, and then I think this is good to merge!

Iinh:
@mdboom Fixed the funny error output! I noticed that the PR checklist "requires a changelog to `HISTORY.rst`, or an explanation of why it does not need one"  - do I need one for this PR? Thanks!

mdboom (approved):
This looks great.  Thanks so much.

Yes -- let's add an entry to `HISTORY.rst`, and then I think this is good to merge.

Iinh:
@mdboom Change log added. Thank you so much for mentoring me through this issue!

PR #263, opened 2021-01-14
Bug 1686481: Version the metrics.yaml and pings.yaml schemas

mdboom:


tests/test_parser.py
```diff
@@ -597,3 +596,31 @@ def test_custom_expires():
         # Double-check that parsing without custom functions breaks
         all_metrics = parser.parse_objects(contents)
         errors = list(all_metrics)
+
+
+def test_historical_versions():
```
wlach, line 601: Probably worth adding a comment or two about what we're trying to test here.

wlach (commented):
This looks pretty reasonable, might be a good idea to get someone more familiar with the parser code base than me to double check this one.

glean_parser/parser.py
```diff
@@ -86,7 +84,14 @@ def _load_file(
...
 
-    filetype = FILE_TYPES.get(schema_key)
+    filetype: Optional[str] = None
+    try:
+        filetype = schema_key.split("/")[-2]
+    except IndexError:
+        filetype = None
+
+    if filetype not in ("metrics", "pings"):
+        filetype = None
```
Dexterp37, line 94: Should the filetype be decided based on the schema, instead of hardcoding names? I know that this behaviour wasn't introduced here, so we can leave it be for now. Maybe we could file a bug for it?

glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -0,0 +1,591 @@
...
+
+---
+$schema: http://json-schema.org/draft-07/schema#
+title: Metrics
+description: |
+  Schema for the metrics.yaml files for Mozilla's Glean telemetry SDK.
+
+  The top-level of the `metrics.yaml` file has a key defining each category of
+  metrics. Categories must be snake_case, and they may also have dots `.` to
+  define subcategories.
```
Dexterp37, line 13: Is it worth capturing the key differences with the previous schema as part of the description? (I'm not sure)

Dexterp37 (approved):
r+wc

Our of curiosity, how are we planning to migrate products to the new version? Glinter warnings/errors?

mdboom:
> Our of curiosity, how are we planning to migrate products to the new version? Glinter warnings/errors?

Good point.  We could add a linter warning "Hey, you're on an old version..."

glean_parser/schemas/metrics.2-0-0.schema.yaml
mdboom: That's a good idea.  The key differences are in the `HISTORY.rst` entry, but documenting the differences directly in the new schemas seems like it would also be useful.

glean_parser/parser.py
mdboom: Yeah -- I think the `$schema` tag should define what's in the file, not hardcoded filenames -- that's how the JSON schema spec says things should work, and also, I like keeping the possibility open of having two `metrics.yaml` files in the same directory, even though none of our users seem to currently do that.

Dexterp37: Right! We weirdly do that for project EXTRACT: I think we use two metrics file there :o How is that even working?!

mdboom: Oh -- to be clear, the current code is based on the schema -- it pulls out the `metrics` part from `moz://mozilla.org/schemas/glean/metrics/2-0-0`.  I forgot EXTRACT was already relying on this.  Another good reason to keep this behavior.

PR #274, opened 2021-02-09
Bug 1691365 - Add a Javascript template to support Glean.js

brizental:

I have tested the Glean.js samples (qt and web extensions) to use this template and it works. 

Will open a PR on the Glean.js repository ~shortly~ tomorrow with the updates.

Example generated file:

```js
/* eslint-disable */

"use strict";

// Global variable to use when in Qt/QML environments,
// where do not have access to the global object (i.e. window);
var pings;

// UMD template based on: https://github.com/umdjs/umd/blob/master/templates/returnExports.js
(function (root, factory) {
    if (typeof define === "function" && define.amd) {
        // AMD. Register as an anonymous module.
        define(["Glean"], factory);
    } else if (typeof module === "object" && module.exports) {
        // Node. Does not work with strict CommonJS, but
        // only CommonJS-like environments that support module.exports, like Node.
        module.exports = factory(require("glean"));
    } else if (typeof root === "undefined") {
        // In Qt/QML environments we can't change the global object from Javascript.
        // We will simply assing to a global variable in this case.
        pings = factory(Glean.Glean);
    } else {
        // Browser globals (root is window)
        if (!root["gleanAssets"]) {
            root["gleanAssets"] = {};
        }
        root["gleanAssets"]["pings"] = factory(root.Glean);
    }
})(typeof self !== "undefined" ? self : this, function(Glean) {
    return {
        /**
        * A custom ping to help showcase Glean.js.
        * It is sent everytime the user clicks the "Submit ping" button
        * on this sample app.
        *
        * Generated from `custom`.
        */
        custom: new Glean._private.PingType({
            includeClientId: true,
            sendIfEmpty: false,
            name: "custom",
            reasonCodes: [],
        }, ),
    };
});
```


HISTORY.rst
```diff
@@ -5,6 +5,8 @@ History
 Unreleased
 ----------
 
+* Add support for generating Javascript code.
```
mdboom, line 8: Should we call out that this is specifically for glean.js (to avoid confusion with Javascript-in-FOG?)

glean_parser/util.py
```diff
@@ -401,20 +401,18 @@ def remove_output_params(d, output_params):
...
+common_metric_args = [
     "category",
     "name",
     "send_in_pings",
     "lifetime",
     "disabled",
+]
+
+
+# Names of paraters that only apply to some of the metrics types.
```
mdboom, line 415:
```suggestion
# Names of parameters that only apply to some of the metrics types.
```

mdboom (approved):
:tada:

chutten (approved):
Dangit, mdboom caught all the same things I was going to catch.

tests/test_javascript.py
```diff
@@ -0,0 +1,193 @@
...
+        name="custom",
+        description="description...",
+        include_client_id=True,
+        bugs=["http://bugzilla.mozilla.com/12345"],
+        notification_emails=["nobody@nowhere.com"],
+    )
+    assert javascript.class_name(ping.type) == "Glean._private.PingType"
+
+
+# TODO: Activate once Glean.js adds support for labeled metric types in Bug 1682573.
```
Dexterp37, line 113: Please add a comment to that bug referring to this comment :)

Dexterp37 (approved):

PR #275, opened 2021-02-10
bug 1685406 - Add telemetry_mirror optional field to metrics schema

chutten:

- [ x] **Quality**: This PR builds and tests run cleanly
- [x ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x ] **Changelog**: This PR includes a changelog entry to `HISTORY.rst` or an explanation of why it does not need one

### DIFF
--- a/glean_parser/metrics.py
+++ b/glean_parser/metrics.py
@@ -57,6 +57,7 @@ class Metric:
         no_lint: Optional[List[str]] = None,
         data_sensitivity: Optional[List[str]] = None,
         defined_in: Optional[Dict] = None,
+        telemetry_mirror: Optional[str] = None,
         _config: Dict[str, Any] = None,
         _validated: bool = False,
     ):
@@ -90,6 +91,8 @@ class Metric:
                 getattr(DataSensitivity, x) for x in data_sensitivity
             ]
         self.defined_in = defined_in
+        if telemetry_mirror is not None:
+            self.telemetry_mirror = telemetry_mirror
 
         # _validated indicates whether this metric has already been jsonschema
         # validated (but not any of the Python-level validation).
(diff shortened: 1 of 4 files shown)


mdboom (changes requested):
I think this also requires changes on the `Metrics.__init__` to store the value on the object, and then probably a small test update by adding this field to one of the example `metrics.yaml` files...

Also I assume (haven't checked the design doc) this is only applicable to certain metric types.  We should restrict that [using something like this](https://github.com/mozilla/glean_parser/blob/5d3db8e3fa3758ad24b93dc09aaef8dbd5571bbb/glean_parser/schemas/metrics.2-0-0.schema.yaml#L521) as well.

chutten:
> I think this also requires changes on the `Metrics.__init__` to store the value on the object, and then probably a small test update by adding this field to one of the example `metrics.yaml` files...
> 
> Also I assume (haven't checked the design doc) this is only applicable to certain metric types. We should restrict that [using something like this](https://github.com/mozilla/glean_parser/blob/5d3db8e3fa3758ad24b93dc09aaef8dbd5571bbb/glean_parser/schemas/metrics.2-0-0.schema.yaml#L521) as well.

It applies to everything except Labeled String (and that's due to a gap in Telemetry's implementation) and JWE. I'm happy to enforce the types lining up properly in the FOG part of the build.

glean_parser/metrics.py
```diff
@@ -90,6 +91,8 @@ def __init__(
                 getattr(DataSensitivity, x) for x in data_sensitivity
             ]
         self.defined_in = defined_in
+        if telemetry_mirror is not None:
```
mdboom, line 94: Should probably just remove the `if` statement so the member is always there, it just might be `None`.  That prevents downstream code from needing to write `if hasattr(metric, "telemetry_mirror")`...

mdboom (approved):
r+wc

glean_parser/metrics.py
mdboom: As you rightly pointed out, this won't work (for schema validation reasons), so seems fine...

PR #281, opened 2021-02-17
Bug 1693323 - Leverage glean_namespace to specify glean environment in JS

brizental:

Have also tested this in the sample Glean.js webextension and it worked. This time I tested with the published @mozilla/glean package.

Sample generated file for reference:

```js
/* eslint-disable */

"use strict";

// Global variable to use when in Qt/QML environments,
// where do not have access to the global object (i.e. window);
var pings;

// Universal Module Definition (UMD) template based on:
// https://github.com/umdjs/umd/blob/master/templates/returnExports.js
(function (root, factory) {
    if (typeof define === "function" && define.amd) {
        // AMD. Register as an anonymous module.
        define(["@mozilla/glean/webext"], factory);
    } else if (typeof module === "object" && module.exports) {
        // Node. Does not work with strict CommonJS, but
        // only CommonJS-like environments that support module.exports, like Node.
        module.exports = factory(require("@mozilla/glean/webext"));
    } else if (typeof root === "undefined") {
        // In Qt/QML environments we can't change the global object from Javascript.
        // We will simply assing to a global variable in this case.
        pings = factory(webext.Glean);
    } else {
        // Browser globals (root is window)
        if (!root["gleanAssets"]) {
            root["gleanAssets"] = {};
        }
        root["gleanAssets"]["pings"] = factory(root.webext);
    }
})(typeof self !== "undefined" ? self : this, function(Glean) {
    return {
        /**
        * A custom ping to help showcase Glean.js
        * It is sent everytime the user clicks the "Submit ping" button
        * on this sample extension popup.
        *
        * Generated from `custom`.
        */
        custom: new Glean._private.PingType({
            includeClientId: true,
            sendIfEmpty: false,
            name: "custom",
            reasonCodes: [],
        }, ),
    };
});
```

### DIFF
--- a/glean_parser/javascript.py
+++ b/glean_parser/javascript.py
@@ -67,16 +67,16 @@ def output_javascript(
 
         - `namespace`: The identifier of the global variable to assign to.
                        This will only have and effect for Qt and static web sites.
-                       Default is `GleanAssets`.
-        - `glean_namespace`: The identifier of the global variable Glean is assigned.
-                             Default is `Glean`.
+                       Default is `gleanAssets`.
+        - `glean_namespace`: Which version of the `@mozilla/glean` to import,
+                             options are `webext` or `qt`. Default is `webext`.
     """
 
     if options is None:
         options = {}
 
     namespace = options.get("namespace", "gleanAssets")
-    glean_namespace = options.get("glean_namespace", "Glean")
+    glean_namespace = options.get("glean_namespace", "webext")
 
     template = util.get_jinja2_template(
         "javascript.jinja2",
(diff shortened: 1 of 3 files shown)


glean_parser/javascript.py
```diff
@@ -67,16 +67,16 @@ def output_javascript(
 
         - `namespace`: The identifier of the global variable to assign to.
                        This will only have and effect for Qt and static web sites.
-                       Default is `GleanAssets`.
-        - `glean_namespace`: The identifier of the global variable Glean is assigned.
-                             Default is `Glean`.
+                       Default is `gleanAssets`.
+        - `glean_namespace`: Which version of the `@mozilla/glean` to import,
```
mdboom, line 71:
```suggestion
        - `glean_namespace`: Which version of the `@mozilla/glean` package to import,
```

mdboom (approved):

PR #286, opened 2021-02-22
bug 1645166 - Add support for `rate` metrics

chutten:
Adds three new metric object model types:
* "rate" for rates that own both their numerator and denominator
* "numerator" for rates that have an external denominator
* "denominator" for counters that are external denominators

### DIFF
--- a/glean_parser/translate.py
+++ b/glean_parser/translate.py
@@ -11,6 +11,7 @@ High-level interface for translating `metrics.yaml` into other formats.
 from pathlib import Path
 import os
 import shutil
+import sys
 import tempfile
 from typing import Any, Callable, Dict, Iterable, List, Optional
 
@@ -57,6 +58,39 @@ OUTPUTTERS = {
     "markdown": Outputter(markdown.output_markdown, []),
     "swift": Outputter(swift.output_swift, ["*.swift"]),
 }
+
+
+def transform_metrics(objects):
+    """
+    Transform the object model from one that represents the YAML definitions
+    to one that reflects the type specifics needed by code generators.
+
+    e.g. This will transform a `rate` to be a `numerator` if its denominator is
+    external.
+    """
+    counters = {}
+    numerators_by_denominator: Dict[str, Any] = {}
+    for category_val in objects.values():
+        for metric in category_val.values():
+            fqmn = metric.identifier()
+            if getattr(metric, "type", None) == "counter":
+                counters[fqmn] = metric
+            denominator_name = getattr(metric, "denominator_metric", None)
+            if denominator_name:
+                metric.type = "numerator"
+                numerators_by_denominator.setdefault(denominator_name, [])
+                numerators_by_denominator[denominator_name].append(fqmn)
+
+    for denominator_name, numerator_names in numerators_by_denominator.items():
+        if denominator_name not in counters:
+            print(
+                f"No `counter` named {denominator_name} found to be used as"
+                "denominator for {numerator_names}",
+                file=sys.stderr,
+            )
+            return 1
+        counters[denominator_name].type = "denominator"
+        counters[denominator_name].numerators = numerator_names
 
 
 def translate_metrics(
--- a/tests/test_translate.py
+++ b/tests/test_translate.py
@@ -190,3 +190,39 @@ def test_getting_line_number():
 
     assert pings["custom-ping"].defined_in["line"] == 7
     assert metrics["core_ping"]["seq"].defined_in["line"] == 27
+
+
+def test_rates(tmpdir):
+    def external_translator(all_objects, output_dir, options):
+        assert (
+            all_objects["testing.rates"]["has_external_denominator"].type == "rate"
+        )  # Hasn't yet been transformed
+
+        translate.transform_metrics(all_objects)
+
+        category = all_objects["testing.rates"]
+        assert category["has_internal_denominator"].type == "rate"
+        assert category["has_external_denominator"].type == "numerator"
+        assert (
+            category["has_external_denominator"].denominator_metric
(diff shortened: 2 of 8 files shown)


glean_parser/translate.py
```diff
@@ -109,6 +110,35 @@ def translate_metrics(
...
         return 1
 
+    # We transform certain metrics typenames depending on context in the file.
+    # e.g. A `rate` may instead be a `numerator` if its denominator is external.
+    value = all_objects.value
+    counters = {}
+    numerators_by_denominator: Dict[str, Any] = {}
+    for category_key, category_val in value.items():
+        for metric_name, metric in category_val.items():
+            fqmn = category_key + "." + metric_name
```
mdboom, line 120: For logic deduplication, let's just use `metric.identifier()` here.

glean_parser/translate.py
```diff
@@ -109,6 +110,35 @@ def translate_metrics(
...
+            if getattr(metric, "type", None) == "counter":
+                counters[fqmn] = metric
+            denominator_name = getattr(metric, "denominator_metric", False)
```
mdboom, line 123: Nit: Feels more Pythonic to use `None` to indicate missing rather than `False`.

glean_parser/translate.py
```diff
@@ -109,6 +110,35 @@ def translate_metrics(
...
+            if denominator_name:
+                metric.type = "numerator"
+                if denominator_name not in numerators_by_denominator:
+                    numerators_by_denominator[denominator_name] = []
```
mdboom, line 127: Could be `numerators_by_denominator.setdefault(denominator_name, [])` to avoid the `if`.

glean_parser/translate.py
```diff
@@ -109,6 +110,35 @@ def translate_metrics(
...
+                numerators_by_denominator[denominator_name].append(fqmn)
+
+    for denominator_name, numerator_names in numerators_by_denominator.items():
+        if denominator_name not in counters:
+            print(
+                "No `counter` named {} found to be used as denominator for {}".format(
```
mdboom, line 133: We can use `f` strings: e.g. `f"No counter named {denominator_name} found to be used as denominator for {numerator_names}"`

glean_parser/translate.py
```diff
@@ -109,6 +110,35 @@ def translate_metrics(
     if util.report_validation_errors(all_objects):
         return 1
 
+    # We transform certain metrics typenames depending on context in the file.
```
mdboom, line 113: This seems close but maybe not quite the right place to put this logic.  However, certain translators (e.g. markdown) won't want this -- there it's probably better to represent things directly as in the metrics.yaml and/or payload.  Perhaps the way to do this is to put it in its own function, which is called from all of the language bindings that need the distinction between rates/counters/numerators/denominators, and skip it in the ones that don't...?  Or alternatively an additional bit of metadata on the `Outputter` objects to control this?

In any event, as a nit, I think this new logic should be in its own function to avoid bloating this one...

tests/test_translate.py
```diff
@@ -190,3 +190,34 @@ def test_getting_line_number():
...
+            category["has_external_denominator"].denominator_metric
+            == "testing.rates.the_denominator"
+        )
+        assert category["also_has_external_denominator"].type == "numerator"
+        assert (
+            category["also_has_external_denominator"].denominator_metric
+            == "testing.rates.the_denominator"
+        )
+        assert category["the_denominator"].type == "denominator"
+        print(category["the_denominator"].numerators)
```
mdboom, line 210: Nit: remove debugging print.

tests/test_translate.py
```diff
@@ -190,3 +190,34 @@ def test_getting_line_number():
...
+        ]
+
+    translate.translate_metrics(
+        ROOT / "data" / "rate.yaml",
+        Path(str(tmpdir)),
+        external_translator,
+        [],
+        options={"foo": "bar"},
+        parser_config={"allow_reserved": True},
+    )
```
mdboom, line 223: If my suggestion above, that markdown shouldn't do this, makes sense, let's add a test for that as well.

mdboom (changes requested):

tests/test_translate.py
chutten: I think from context the new patch handles this without needing an explicit test. Be sure to disagree, though, if I'm wrong : )

glean_parser/translate.py
```diff
@@ -113,6 +147,9 @@ def translate_metrics(
     if parser_config.get("allow_reserved"):
         options["allow_reserved"] = True
 
+    # Show the outputter where to find the optional transformer
+    options["transform"] = transform_metrics
```
mdboom, line 151: Ah, sorry, I wasn't clear.  I think if we're going to do this, we might as well just have the outputter call the function directly.  (I think it's a little opaque that's it's passed in the `options` dict...)  That, of course, requires each of the outputters to call this (which now they don't do).

Alternatively (and maybe better because it localizes things) -- we could have a `transform_metrics` flag on the `Outputters` metadata [here](https://github.com/mozilla/glean_parser/blob/c9e4507b36bfe7e27412f2f51866ce1398860dff/glean_parser/translate.py#L44).  This would be `True` for everything except markdown.  Then at this line we would perform metric transformation (or not) based on that flag.

chutten: I like the magic transform thing, but it'd be 100% False for the foreseeable future as I'm implementing `rate` just for FOG and FOG doesn't use `translate` at all.

So, come to think of it, that's a real problem with my design. I guess I'd have to use the exposed `translate.transform` by name _anyway_. Lemme rework to that.

mdboom (approved):

PR #301, opened 2021-04-06
Update schema to allow for "special" glean_* pings

mdboom:
See https://github.com/mozilla/glean/pull/1572

### DIFF
--- a/glean_parser/schemas/metrics.2-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.2-0-0.schema.yaml
@@ -32,7 +32,7 @@ definitions:
   # were also supported.
   kebab_case:
     type: string
-    pattern: "^[a-z][a-z0-9-]{0,29}$"
+    pattern: "^[a-z][a-z0-9-]{0,29}$|^glean_.*$"
 
   long_id:
     allOf:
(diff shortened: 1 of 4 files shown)


glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -32,7 +32,7 @@ definitions:
   # were also supported.
   kebab_case:
     type: string
-    pattern: "^[a-z][a-z0-9-]{0,29}$"
+    pattern: "^[a-z][a-z0-9-]{0,29}$|^glean_.*$"
```
chutten, line 35: I'm not sure how much bikeshed painting you're interested in, but I think what we'd ideally do is have a `ping_or_section_name` definition that has this pattern with a comment that it's to include ping names that aren't actually ping names like `glean_client_info`. 

This will allow us to forbid `glean_` ping definitions in pings.yaml (which can use `kebab_case` directly) while permitting `glean_` in the `send_in_pings` line.

chutten: Either way, we need a comment.

chutten (approved):

glean_parser/schemas/metrics.2-0-0.schema.yaml
mdboom: I can never have too much bikeshed paint!

Good point about this -- I moved the special case to `send_in_pings` itself, and added a comment.

chutten (approved):

PR #310, opened 2021-04-13
Bug 1700991 - Remove support for the JWE metric type

badboy:
I _think_ modifying the schema here is fine. I did that to provide an explicit error message. Does anyone else depend on this schema directly?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Raise limit on number of statically-defined lables to 100. ([bug 1702263](https://bugzilla.mozilla.org/show_bug.cgi?id=1702263))
 - BUGFIX: Version 2.0.0 of the schema now allows the "special" `glean_.*` ping names for Glean-internal use again.
+- Remove support for JWE metric types.
 
 ## 2.5.0 (2021-02-23)
 
--- a/glean_parser/markdown.py
+++ b/glean_parser/markdown.py
@@ -34,9 +34,6 @@ def extra_info(obj: Union[metrics.Metric, pings.Ping]) -> List[Tuple[str, str]]:
         for label in obj.ordered_labels:
             extra_info.append((label, None))
 
-    if isinstance(obj, metrics.Jwe):
-        extra_info.append(("decrypted_name", obj.decrypted_name))
-
     if isinstance(obj, metrics.Quantity):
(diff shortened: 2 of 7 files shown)


badboy:
ah, also: I only modified v2 of the schema. Should we change that in v1 too?

brizental:
> ah, also: I only modified v2 of the schema. Should we change that in v1 too?

Is it even necessary to change the schema at all? Or is [this](https://github.com/mozilla/glean_parser/pull/310/files#diff-7f9916942ae822edeacb72774bb4d3e274a6ffd4774bd9d42403451db77d18a3R328-R331) already enough to deny JWE?

If it's an issue (which I am not sure about) it would be possible to just change the `description` field and leave the `decrypted_name` part alone. This way, users checking the schema (e.g. through IDE tooltips) would see in the description that the field is not supported anymore.

Also, you can mark a field as [`deprecated`](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3) through the schema, it's not really the case for JWE, but could work as well.

badboy:
> > ah, also: I only modified v2 of the schema. Should we change that in v1 too?
> 
> Is it even necessary to change the schema at all? Or is [this](https://github.com/mozilla/glean_parser/pull/310/files#diff-7f9916942ae822edeacb72774bb4d3e274a6ffd4774bd9d42403451db77d18a3R328-R331) already enough to deny JWE?

It should also deny with only that code change. However using the schema for that results in a nicer error message IMO.

> If it's an issue (which I am not sure about) it would be possible to just change the `description` field and leave the `decrypted_name` part alone. This way, users checking the schema (e.g. through IDE tooltips) would see in the description that the field is not supported anymore.

Not really that big of an issue, even just the code change should actively deny it and adjusting the documentation can take care of the rest.
I'm waiting for @mdboom to see if he knows the implications of the schema change.

> Also, you can mark a field as [`deprecated`](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3) through the schema, it's not really the case for JWE, but could work as well.

TIL, thanks, will check.

mdboom (approved):
This looks fine to me.  Making the change in the schema will provide a line number for the error for the user, which you don't get from only changing the class `__init__`.

I think only changing 2.0.0 is fine.

PR #321, opened 2021-04-30
Bug 1693486 - Generate new event API construct

badboy:
When a metric of `type: event` includes any `type` annotation in its
extra keys list, the new object/struct-based API will be used.

Only three types are supported right now:

* string
* boolean
* quantity (an integer)

This is only implemented for Kotlin & Swift.
Python will need this in the Glean Python code base.
Glean.js will need its own implementation first.

FOG will need to implement something similar for its Rust, C++ and
JavaScript API.

---

Needs to land so that Glean can use it here: https://github.com/mozilla/glean/pull/1603

### DIFF
--- a/glean_parser/schemas/metrics.2-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.2-0-0.schema.yaml
@@ -348,6 +348,12 @@ definitions:
           properties:
             description:
               type: string
+            type:
+              type: string
+              enum:
+                - string
+                - boolean
+                - quantity
           required:
             - description
         maxProperties: 10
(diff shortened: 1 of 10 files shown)


glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -348,6 +348,12 @@ definitions:
           properties:
             description:
               type: string
+            type:
```
Dexterp37, line 351: Just to double check: this does not require a schema bump, right?

Dexterp37 (approved):

glean_parser/schemas/metrics.2-0-0.schema.yaml
badboy: It should not, because it's optional!

PR #339, opened 2021-05-25
Add missing parts for new events API

badboy:
### Pull Request checklist ###

- [ ] **Quality**: This PR builds and tests run cleanly
  - `make test` runs without emitting any warnings
  - `make lint` runs without emitting any errors
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry to `CHANGELOG.md` or an explanation of why it does not need one
  - Any breaking changes to language binding APIs are noted explicitly

PR #340, opened 2021-05-27
Don't rely on setup.py for installation anymore

badboy:
Based on #339, only the last commit is relevant.


mdboom (approved):
+1 (but looks like it's combined with commits from another branch)

PR #346, opened 2021-06-10
Bug 1713398 - Enable custom distribution without a gecko_datapoint 

badboy:
Rust doesn't have named arguments, so we need to ensure they are in
order.

### DIFF
--- a/glean_parser/util.py
+++ b/glean_parser/util.py
@@ -420,14 +420,16 @@ common_metric_args = [
 
 
 # Names of parameters that only apply to some of the metrics types.
+# **CAUTION**: This list needs to be in the order the Swift & Rust type constructors
+# expects them. (The other language bindings don't care about the order).
 extra_metric_args = [
     "time_unit",
     "memory_unit",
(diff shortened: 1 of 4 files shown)


glean_parser/util.py
```diff
@@ -420,22 +420,24 @@ def remove_output_params(d, output_params):
...
+# expects them. (The other language bindings don't care about the order).
 extra_metric_args = [
     "time_unit",
     "memory_unit",
     "allowed_extra_keys",
     "reason_codes",
-    "bucket_count",
-    "range_max",
     "range_min",
+    "range_max",
```
Dexterp37, line 431: Uh oh. Does this mean it was broken before?

Dexterp37 (approved):

glean_parser/util.py
badboy: Well, only for Rust and for that it wasn't supported so far.

badboy: (and technically for Swift, but it has no implementation for it either)

PR #355, opened 2021-07-02
Bug 1703577 - Check for redundant words in ping names

jennshan:

- [ ] **Tests**: ### This PR includes thorough tests or an explanation of why it does not

make lint fails with error that says "would reformat glean_parser/lint.py"
Haven't made the tests, but creating a PR to look for feedback on my code

### DIFF
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -42,7 +42,7 @@ def _split_words(name: str) -> List[str]:
     """
     Helper function to split words on either `.` or `_`.
     """
-    return re.split("[._]", name)
+    return re.split("[._-]", name)
 
 
 def _english_list(items: List[str]) -> str:
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -262,6 +262,36 @@ def check_expired_metric(
         yield ("Metric has expired. Please consider removing it.")
 
 
+def check_redundant_ping(
+    pings: Iterable[pings.Ping]
+) -> LintGenerator:
+    '''
+    Check if the pings contains 'ping' as the prefix or suffix, or 'ping' or 'custom'
+    '''
+    ping_words = sorted([_split_words(ping.name) for ping in pings])
+
+    if len(ping_words) != 0:
+        ping_first_word = ping_words[0]
+        ping_last_word = ping_words[-1]
+
+        if ping_first_word == "ping":
+            yield (
+                "The prefix 'ping' is redundant."
+            )
+        elif ping_last_word == "ping":
+            yield (
+                "The suffix 'ping' is redundant."
+            )
+        elif "ping" in ping_words:
+            yield (
+                "The word 'ping' is redundant."
+            )
+        elif "custom" in ping_words:
+            yield (
+                "The word 'custom' is redundant."
+            )
+
+
 # The checks that operate on an entire category of metrics:
 #    {NAME: (function, is_error)}
 CATEGORY_CHECKS: Dict[
@@ -281,6 +311,7 @@ METRIC_CHECKS: Dict[
     "BUG_NUMBER": (check_bug_number, CheckType.error),
     "BASELINE_PING": (check_valid_in_baseline, CheckType.error),
     "MISSPELLED_PING": (check_misspelled_pings, CheckType.error),
+    "REDUNDANT_PING": (check_redundant_ping, CheckType.error),
     "EXPIRATION_DATE_TOO_FAR": (check_expired_date, CheckType.warning),
     "USER_LIFETIME_EXPIRATION": (check_user_lifetime_expiration, CheckType.warning),
     "EXPIRED": (check_expired_metric, CheckType.warning),
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -263,33 +263,25 @@ def check_expired_metric(
 
 
 def check_redundant_ping(
-    pings: Iterable[pings.Ping]
+    pings: pings.Ping, parser_config: Dict[str, Any]
 ) -> LintGenerator:
-    '''
+    """
     Check if the pings contains 'ping' as the prefix or suffix, or 'ping' or 'custom'
-    '''
-    ping_words = sorted([_split_words(ping.name) for ping in pings])
+    """
+    ping_words = _split_words(pings.name)
 
     if len(ping_words) != 0:
         ping_first_word = ping_words[0]
         ping_last_word = ping_words[-1]
 
         if ping_first_word == "ping":
-            yield (
-                "The prefix 'ping' is redundant."
-            )
+            yield ("The prefix 'ping' is redundant.")
         elif ping_last_word == "ping":
-            yield (
-                "The suffix 'ping' is redundant."
-            )
+            yield ("The suffix 'ping' is redundant.")
         elif "ping" in ping_words:
-            yield (
-                "The word 'ping' is redundant."
-            )
+            yield ("The word 'ping' is redundant.")
         elif "custom" in ping_words:
-            yield (
-                "The word 'custom' is redundant."
-            )
+            yield ("The word 'custom' is redundant.")
 
 
 # The checks that operate on an entire category of metrics:
@@ -311,7 +303,6 @@ METRIC_CHECKS: Dict[
     "BUG_NUMBER": (check_bug_number, CheckType.error),
     "BASELINE_PING": (check_valid_in_baseline, CheckType.error),
     "MISSPELLED_PING": (check_misspelled_pings, CheckType.error),
-    "REDUNDANT_PING": (check_redundant_ping, CheckType.error),
     "EXPIRATION_DATE_TOO_FAR": (check_expired_date, CheckType.warning),
     "USER_LIFETIME_EXPIRATION": (check_user_lifetime_expiration, CheckType.warning),
     "EXPIRED": (check_expired_metric, CheckType.warning),
@@ -324,6 +315,7 @@ PING_CHECKS: Dict[
     str, Tuple[Callable[[pings.Ping, dict], LintGenerator], CheckType]
 ] = {
     "BUG_NUMBER": (check_bug_number, CheckType.error),
+    "REDUNDANT_PING": (check_redundant_ping, CheckType.error),
 }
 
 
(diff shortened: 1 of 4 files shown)


glean_parser/lint.py
```diff
@@ -281,6 +311,7 @@ def check_expired_metric(
     "BUG_NUMBER": (check_bug_number, CheckType.error),
     "BASELINE_PING": (check_valid_in_baseline, CheckType.error),
     "MISSPELLED_PING": (check_misspelled_pings, CheckType.error),
+    "REDUNDANT_PING": (check_redundant_ping, CheckType.error),
```
badboy, line 314: This will need to go further below, around line 322, as we only want that check running on `pings`, not `metrics`.

glean_parser/lint.py
```diff
@@ -262,6 +262,36 @@ def check_expired_metric(
         yield ("Metric has expired. Please consider removing it.")
 
 
+def check_redundant_ping(
+    pings: Iterable[pings.Ping]
```
badboy, line 266: This check will receive one ping object at a time, so this should be `ping: pings.Ping`.
Additionally we always pass in the configuration object as well, so it takes a second argument.
The full thing should read

```
def check_redundant_ping(
    ping: pings.Ping, parser_config: Dict[str, Any]
) -> LintGenerator:
```

glean_parser/lint.py
```diff
@@ -262,6 +262,36 @@ def check_expired_metric(
...
+) -> LintGenerator:
+    '''
+    Check if the pings contains 'ping' as the prefix or suffix, or 'ping' or 'custom'
+    '''
+    ping_words = sorted([_split_words(ping.name) for ping in pings])
```
badboy, line 271: With only one ping being passed in this can become:
```suggestion
    ping_words = _split_words(ping.name)
```

glean_parser/lint.py
```diff
@@ -262,6 +262,36 @@ def check_expired_metric(
         yield ("Metric has expired. Please consider removing it.")
 
 
+def check_redundant_ping(
+    pings: Iterable[pings.Ping]
+) -> LintGenerator:
+    '''
```
badboy, line 268: This should use double quotes like `"""`

glean_parser/lint.py
```diff
@@ -262,6 +262,36 @@ def check_expired_metric(
...
+    ping_words = sorted([_split_words(ping.name) for ping in pings])
+
+    if len(ping_words) != 0:
+        ping_first_word = ping_words[0]
+        ping_last_word = ping_words[-1]
+
+        if ping_first_word == "ping":
+            yield (
+                "The prefix 'ping' is redundant."
+            )
```
badboy, line 280: Best to run `make lint`, which will also reformat this (and following lines) to

```
yield ("The prefix 'ping' is redundant.")
```

(to reduce vertical space)

badboy (changes requested):
This looks good to start. I left some comment inline that I spotted already.
With these fixed you should be able to run the current tests.
You can then proceed to add a new test for this lint check explicitly.

glean_parser/lint.py
jennshan: I see, now I placed the
"REDUNDANT_PING": (check_redundant_ping, CheckType.error),

under the PING_CHECKS: Dict[..]

glean_parser/lint.py
jennshan: I didn't understood the use of the second argument that is why I didn't had, now that you said we always pass this argument I included it.

glean_parser/lint.py
jennshan: Got you!

glean_parser/lint.py
badboy: It's not used in your code, but because of how we call these functions it's always passed in case the function _does_ need it.

glean_parser/lint.py
jennshan: Done

jennshan:
Hi, 

Added test_redundant_pings function to tests/test_lint.py, please review and comment

glean_parser/lint.py
```diff
@@ -262,6 +262,28 @@ def check_expired_metric(
         yield ("Metric has expired. Please consider removing it.")
 
 
+def check_redundant_ping(
+    pings: pings.Ping, parser_config: Dict[str, Any]
+) -> LintGenerator:
+    """
+    Check if the pings contains 'ping' as the prefix or suffix, or 'ping' or 'custom'
+    """
+    ping_words = _split_words(pings.name)
```
badboy, line 271: Unfortunately pings can contain dashes (`-`) and `_split_words` doesn't split on that.
Luckily it should be easy to fix: add the `-` to the function like this:

```
def _split_words(name: str) -> List[str]:
    """
    Helper function to split words on either `.` or `_`.
    """
    return re.split("[._-]", name)
```

This should not break anything further.

tests/test_lint.py
```diff
@@ -336,3 +336,27 @@ def test_bug_number_pings(content, num_nits):
     assert len(nits) == num_nits
     if num_nits > 0:
         assert set(["BUG_NUMBER"]) == set(v.check_name for v in nits)
+
+
+def test_redundant_pings():
```
badboy, line 341: The test above this uses `ping` as the ping name. Change that to `search` (line 317ff).

tests/test_lint.py
```diff
@@ -336,3 +336,27 @@ def test_bug_number_pings(content, num_nits):
...
+    content = [
+        {
+            "ping_test": {
+                "description": "testing the ping in the prefix",
+                "send_if_empty": True,
+                "bugs": ["https://bugzilla.mozilla.org/show_bug.cgi?id=1703577"],
+                "data_review": ["https://bugzilla.mozilla.org/show_bug.cgi?id=1703577"],
+            }
+        }
+    ]
```
badboy, line 354: This right now fails because `content` needs to be an object, not a list.
Additionally it turns out `add_required_ping` adds most of these fields for us.
So this becomes:

```suggestion
    content = {
        "ping": {}
    }
```

If you really want to expand this test you can parametrize that test just as the one above, using `@pytest.mark.parametrize`.
This allows you to specify a list of inputs this test function is run with.

This is fully optional though, it's also ok if you leave the test barebones for now.

badboy (changes requested):
We're nearly there.

There's pings in the file `tests/data/pings.yaml` that contain `ping` or `custom`. You can add the line `no_lint: [REDUNDANT_PING]` to the 3 pings there.

And last but not least please add a line to the `CHANGELOG.md`:

```
- New lint: Check for redundant words in ping names ([#355](https://github.com/mozilla/glean_parser/pull/355))

File: glean_parser/lint.py (hunk 6 above)
Comment (jennshan) [reply glean_parser/lint.py:271]: I see, sure I will include '-' to the built-in function split

File: tests/test_lint.py [hunk 9]
```diff
@@ -336,3 +336,27 @@ def test_bug_number_pings(content, num_nits):
...
+    content = util.add_required_ping(content)
+    all_pings = parser.parse_objects([content])
+    errs = list(all_pings)
+    assert len(errs) == 0
+
+    nits = lint.lint_metrics(all_pings.value)
+    assert len(nits) == 1
+    assert set(["REDUNDANT_PING"]) == set(v.check_name for v in nits)
```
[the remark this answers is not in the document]
jennshan: Done

tests/test_lint.py
jennshan: Changed, the name of all the three above three test pings from 'ping' to 'search'

tests/test_lint.py
jennshan: I see, I changed `content` type from list to object and added an barebone `ping` object

tests/test_lint.py
```diff
@@ -336,3 +336,21 @@ def test_bug_number_pings(content, num_nits):
...
         assert set(["BUG_NUMBER"]) == set(v.check_name for v in nits)
+
+
+def test_redundant_pings():
+    """
+    Test that name contains '-ping' or 'ping-' or 'ping' or 'custom' yields lint errors.
+    """
+    content = {
+            "ping": {}
+    }
```
badboy, line 347: The linter wants this reformatted.

```suggestion
    content = {"ping": {}}
```

(I really need to change CI to still run the tests regardless of lint failures)

badboy (changes requested):
One last thing and then this should be done!

tests/test_lint.py
jennshan: How to catch these errors myself, because `make lint` was successful locally

badboy:
```
❯ make lint
python3 -m flake8 glean_parser tests
python3 -m black --check glean_parser tests setup.py
would reformat tests/test_lint.py
Oh no! 💥 💔 💥
1 file would be reformatted, 30 files would be left unchanged.
make: *** [lint] Error 1
```

It does fail the lint test, but it's not perfectly obvious.
In #357 I did change it so that it shows the difference:

```
❯ make lint
python3 -m flake8 glean_parser tests
python3 -m black --check --diff glean_parser tests setup.py
--- tests/test_lint.py	2021-07-08 08:50:20.282340 +0000
+++ tests/test_lint.py	2021-07-08 08:50:49.756407 +0000
@@ -340,13 +340,11 @@

 def test_redundant_pings():
     """
     Test that name contains '-ping' or 'ping-' or 'ping' or 'custom' yields lint errors.
     """
-    content = {
-            "ping": {}
-    }
+    content = {"ping": {}}

     content = util.add_required_ping(content)
     all_pings = parser.parse_objects([content])
     errs = list(all_pings)
     assert len(errs) == 0
would reformat tests/test_lint.py
Oh no! 💥 💔 💥
1 file would be reformatted, 30 files would be left unchanged.
make: *** [lint] Error 1
```

That should now be more obvious.
You can even auto-format the files again:

```
❯ python3 -m black glean_parser tests setup.py
reformatted tests/test_lint.py
All done! ✨ 🍰 ✨
1 file reformatted, 30 files left unchanged.
```

We currently don't have a make task for that, but I should add it.

badboy: Done: #358

badboy (approved):

PR #373, opened 2021-08-17
Bug 1693487 - Define interfaces for event extras on Typescript template

brizental:

  - Any breaking changes to language binding APIs are noted explicitl
 
Example:

```ts

// AUTOGENERATED BY glean_parser. DO NOT EDIT. DO NOT COMMIT.

import EventMetricType from "@mozilla/glean/webext/private/metrics/event";

/**
 * Just testing events
 *
 * Generated from `core.event_example`.
 */
export const eventExample = new EventMetricType<{
    key1?: string,
    key2?: string,
}>({
    category: "core",
    name: "event_example",
    sendInPings: ["events"],
    lifetime: "ping",
    disabled: false,
}, ["key1", "key2"]);

/**
 * Just testing events
 *
 * Generated from `core.preference_toggled`.
 */
export const preferenceToggled = new EventMetricType<{
    enabled?: boolean,
    preference?: string,
    swapped?: number,
}>({
    category: "core",
    name: "preference_toggled",
    sendInPings: ["events"],
    lifetime: "ping",
    disabled: false,
}, ["enabled", "preference", "swapped"]);
```

This PR is a companion to https://github.com/mozilla/glean.js/pull/630


glean_parser/metrics.py
```diff
@@ -312,7 +312,16 @@ def allowed_extra_keys(self):
     def allowed_extra_keys_with_types(self):
         # Sort keys so that output is deterministic
         return sorted(
-            [(k, v["type"]) for (k, v) in self.extra_keys.items()], key=lambda x: x[0]
+            [(k, v["type"]) for (k, v) in self.extra_keys.items() if "type" in v],
```
Dexterp37, line 315: What happens if `type` is not in `v`? Do we simply skip it or is it accounted for due to `allowed_extra_keys_without_types`? I'm not sure that would ever happen, right?

badboy, line 315: Hm, I didn't consider the case where only some of the extra keys have a type set. Edge case, but possible I guess.

We can simplify it though and set a default type here as a fallback:

```suggestion
            [(k, v.get("type", "string")) for (k, v) in self.extra_keys.items() if "type" in v],
```

Then you don't need the `_without_types` method below and the template won't need the extra call either.

badboy: (plus this makes it work for the other languages too)

badboy (changes requested):

glean_parser/metrics.py
brizental: > What happens if type is not in v? 

This function was failing when `type` was not in `v`. I guess it was not called anywhere else, that is why it was not causing trouble before. With the current approach we simply skip it, but jan-erik's comment changes it and is nicer than the current approachm heh so I'll apply it.

badboy (approved):

PR #374, opened 2021-08-18
Bug 1726326 - Add support for the text metric type

badboy:
This applies the following additional restrictions:

1) Metrics of this type will need to be data collection category 3 or 4.
2) Metrics of this type will only be allowed for custom pings. Sending a text metric in built-in pings is forbidden.
3) Metrics of this type will only be allowed to have a ping or application lifetime. Other lifetimes will be forbidden.

Restrictions are applied through the jsonschema.
This leads to okayish error messages for 1) and 3)
and a suboptimal, but understandable error message for 2).

---

See this output:

```
❯ python -m glean_parser -- translate --output tmp --format kotlin tests/data/text_invalid.yaml
==============================================================================
/Users/jer/mozilla/src/glean_parser/tests/data/text_invalid.yaml:
    ```
    invalid.text:
      lifetime:
        lifetime: user
    ```

    'user' is not one of ['ping', 'application']

    Documentation for this node:
        Text metrics must have ping or application lifetime.

==============================================================================
/Users/jer/mozilla/src/glean_parser/tests/data/text_invalid.yaml:
    ```
    invalid.text:
      sensitivity:
        data_sensitivity:
        - technical
    ```

    'technical' is not one of ['web_activity', 'highly_sensitive']
==============================================================================
/Users/jer/mozilla/src/glean_parser/tests/data/text_invalid.yaml:
    ```
    invalid.text:
      builtin_pings:
        send_in_pings:
        - metrics
    ```

    OrderedDict([('description', 'Text metrics can only be sent in custom
    pings. Built-in pings are not allowed.'), ('pattern',
    '^(metrics|baseline|events|deletion-request|default|glean_.*)$')]) is
    not allowed for 'metrics'
ERROR running glean_parser v3.7.1.dev3+g0f92902.d20210809
```

### DIFF
--- a/glean_parser/schemas/metrics.2-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.2-0-0.schema.yaml
@@ -117,6 +117,8 @@ definitions:
 
                 `labeled_boolean`, `labeled_string`, `labeled_counter`.
 
+            - `text`: Record long text data.
+
         type: string
         enum:
           - event
@@ -137,6 +139,7 @@ definitions:
           - labeled_string
           - labeled_counter
           - rate
+          - text
 
       description:
         title: Description
@@ -625,3 +628,41 @@ additionalProperties:
               description: |
                 `denominator_metric` is only allowed for `rate`.
               maxLength: 0
+      -
+        if:
+          properties:
+            type:
+              const: text
+        then:
+          properties:
+            lifetime:
+              description: >
+                Text metrics must have ping or application lifetime.
+              enum:
+                - ping
+                - application
+
+            data_sensitivity:
+              description: >
+                Text metrics require Category 3 (`web_activity`)
+                or Category 4 (`highly_sensitive`).
+              type: array
+              items:
+                enum:
+                  - web_activity
+                  - highly_sensitive
+
+            send_in_pings:
+              description: |
+                Text metrics can only be sent in custom pings.
+                Built-in pings are not allowed.
+              type: array
+              items:
+                allOf:
+                  - $ref: "#/definitions/kebab_case"
+                  - not:
+                      description: >
+                        Text metrics can only be sent in custom pings.
+                        Built-in pings are not allowed."
+                      pattern:
+                        "^(metrics|baseline|events|deletion-request|default|glean_.*)$"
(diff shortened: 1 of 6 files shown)


glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -625,3 +628,36 @@ additionalProperties:
...
+            send_in_pings:
+              description: |
+                Text metrics can only be sent in custom pings. Built-in pings are not allowed.
+              type: array
+              items:
+                allOf:
+                  - $ref: "#/definitions/kebab_case"
+                  - not:
+                      description: "Text metrics can only be sent in custom pings. Built-in pings are not allowed."
+                      pattern: "^(metrics|baseline|events|deletion-request|default|glean_.*)$"
```
brizental, line 663: Hum, TIL about `default`. Is that a keyword someone can add in order to get the default ping for the metric type? If so it should be documented [here](https://mozilla.github.io/glean/book/reference/yaml/metrics.html#send_in_pings) and should also be disallowed as a ping name (I just checked that it is not).

These are all follow up stuff though.

brizental (approved):

glean_parser/schemas/metrics.2-0-0.schema.yaml
badboy: hm, I think so? https://github.com/mozilla/glean_parser/blob/text-metric/glean_parser/metrics.py#L80-L81

brizental: Yeah it is also mentioned [here](https://mozilla.github.io/glean/book/user/pings/custom.html#sending-metrics-in-a-custom-ping) (found this after sending the review). Either way, good to document it in the metrics parameters as well.

badboy: https://github.com/mozilla/glean_parser/blob/text-metric/glean_parser/parser.py#L337-L340

brizental: > https://github.com/mozilla/glean_parser/blob/text-metric/glean_parser/parser.py#L337-L340

Yeah, this checks the `send_in_pings` parameter, but the ping name is not checked. Currently `default` is an accepted ping name. Of course this is a minor bug, but still.

badboy: Right, I'll fix that separately.

PR #411, opened 2021-11-04
Bug 1739126: Add linter about use of private bugtracker

mdboom:

### DIFF
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -18,6 +18,7 @@ from typing import (
     Tuple,
     Union,
 )  # noqa
+from urllib.parse import urlparse
 
 
 from . import metrics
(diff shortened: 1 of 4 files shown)


glean_parser/lint.py
```diff
@@ -273,6 +274,18 @@ def check_expired_metric(
         yield ("Metric has expired. Please consider removing it.")
 
 
+PRIVATE_BUGTRACKER_DISALLOW_LIST = set(["mozilla-hub.atlassian.net"])
```
mdboom, line 277: I waffled a lot about allow vs. disallow list.  Thoughts?  I don't think actually checking the URLs at parse time is feasible -- for one, glean using products may be built in an offline environment...

wlach: I think an allowlist might be better, as that allows us to encode a tighter policy (and I imagine, hypothetically, others using Glean might want to do the same). In the case of most Mozilla products, IMO the allowed URLs should be bugzilla.mozilla.org or github.

glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -230,9 +230,11 @@ definitions:
...
           **Required.**
 
-          A list of bug URLs (e.g. Bugzilla and Github) that are relevant to
-          this metric, e.g., tracking its original implementation or later
-          changes to it.
+          A list of bug URLs that are relevant to this metric, e.g., tracking
+          its original implementation or later changes to it.
+
+          This URL should be openly visible, and not behind authorization. (e.g.
+          a public Github repository or Bugzilla, not a private JIRA instance).
```
wlach, line 237: We should also document this in the glean documentation e.g. https://mozilla.github.io/glean/book/reference/yaml/metrics.html#bugs

wlach (changes requested):
Everything looks in order except I think we might want to use an allowlist rather than a disallowlist. Let me know what you think.

A future enhancement would be to make either of these configurable

glean_parser/lint.py
mdboom: I had considered that.  Github is a tricky one -- it could be public or private and I don't think there's any way to know from the URL.  I guess that problem exists whether its an allow or disallow list, though.

PR #425, opened 2021-12-23
Bug 1677434 - Enable Rust code generation from the glean_parser

travis79:
This is a draft because I still need to test this in a real project


travis79:
I've got a basic proof of concept integration with Nimbus, but it requires me to have a python env set up with glean_parser (the correct one) in it. I'm not quite sure what is the best way to get glean_parser and it's dependencies set up from the `build.rs` build-script.

ref: https://github.com/mozilla/application-services/pull/4759

glean_parser/rust.py
```diff
@@ -0,0 +1,199 @@
...
+                for subvalue in list(value):
+                    if not first:
+                        yield ", "
+                    yield from self.iterencode(subvalue)
+                    first = False
+                yield "]"
+            elif value is None:
+                yield "None"
+            elif isinstance(value, str):
+                yield '"' + value + '".into()'
```
mdboom, line 64:
```suggestion
                yield f'"{value}.into()"'
```

glean_parser/rust.py
```diff
@@ -0,0 +1,199 @@
...
+    if obj_type.startswith("labeled_"):
+        obj_type = obj_type[8:]
+    return util.Camelize(obj_type) + "Metric"
+
+
+def extra_keys(allowed_extra_keys):
+    """
+    Returns the &'static [&'static str] ALLOWED_EXTRA_KEYS for impl ExtraKeys
+    """
+    return "&[" + ", ".join(map(lambda key: '"' + key + '"', allowed_extra_keys)) + "]"
```
mdboom, line 134:
```suggestion
    return "&[" + ", ".join([f'"{key}"' for key in allowed_extra_keys]) + "]"
```

glean_parser/rust.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class Category:
```
mdboom, line 137: Maybe slightly better to use a constructor (`__init__`) here rather than setting the methods directly.  (Practically, since this is kind of a private class it doesn't matter much, but it would if this were a public API).  See https://github.com/mozilla/glean_parser/blob/main/glean_parser/metrics.py#L36 for an example.

glean_parser/templates/rust.jinja2
```diff
@@ -0,0 +1,287 @@
...
+{% for typ, metrics in metric_by_type.items() %}
+{% if typ.0 in ('BOOLEAN_MAP', 'COUNTER_MAP', 'STRING_MAP') %}
+        pub static {{typ.0}}: Lazy<RwLock<HashMap<MetricId, Labeled{{typ.1}}>>> = Lazy::new(||
+            RwLock::new(HashMap::new())
+        );
+{% endif %}
+{% endfor%}
+    }
+}
+{% endif %}
```
mdboom, line 287: Wow.  That's a lot of jinja.

mdboom (approved):
Looks good.  This definitely seems like the most involved one of these we've done.

My comments are just to move to more modern Python idioms.  If you think they decrease readability (because they are less obvious for a Python novice), feel free to ignore -- none of my suggestions change logic or performance or anything...

PR #432, opened 2022-01-07
Reject files if they use `unit` by mistake

badboy:
The time metric types use `time_unit`, not `unit`.
Easy mistake to made, we can catch that.
Note: `time_unit` is OPTIONAL for all three metric types.

This is not a problem for `memory_distribution`, because its
`memory_unit` is REQUIRED.

### DIFF
--- a/glean_parser/schemas/metrics.2-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.2-0-0.schema.yaml
@@ -434,10 +434,14 @@ definitions:
       unit:
         title: Unit
         description: |
-          The unit of the metric, for metrics that don't already require a
-          meaningful unit, such as `time_unit`.
+          The unit of the metric.
+          This is only required for metrics
+          that don't already require a meaningful unit, e.g. `quantity`
           This is provided for informational purposes only and doesn't have any
           effect on data collection.
+
+          Metric types like `timespan`, `datetime`
+          and `timing_distribution` take a `time_unit` instead.
         type: string
 
       no_lint:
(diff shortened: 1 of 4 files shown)


glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -435,7 +435,7 @@ definitions:
         title: Unit
         description: |
           The unit of the metric, for metrics that don't already require a
-          meaningful unit, such as `time_unit`.
+          meaningful unit, such as `quantity`.
```
chutten, line 438: `time_unit` is the meaningful unit. But I agree this wording could be updated.

glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -685,3 +685,20 @@ additionalProperties:
...
+      -
+        if:
+          properties:
+            type:
+              enum:
+                - datetime
+                - timing_distribution
+                - timespan
+          required:
+            - unit
```
chutten, line 698: Oh wow, I had to read the note under [this part of the reference](https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else) to understand what's going on here.

So, if the type is one of the three _and_ there is a `unit`, _then_ we require a `time_unit` with a helpful message.

Which suggests to me that this will fail to error if someone includes both a `unit` _and_ a `time_unit` field? We may need a test for that.

glean_parser/schemas/metrics.2-0-0.schema.yaml
badboy: Woops. accidental change.
But yes, re-reading that it's hard to understand.

glean_parser/schemas/metrics.2-0-0.schema.yaml
mdboom: Yeah, it definitely took my brain a little while to understand what this is doing.  Maybe a comment?

tests/test_cli.py
```diff
@@ -155,3 +155,19 @@ def test_reject_jwe(tmpdir):
...
+        __main__.main,
+        [
+            "glinter",
+            str(ROOT / "data" / "wrong_key.yamlx"),
+        ],
+    )
+    assert result.exit_code == 1
+    # wrong `unit` key for datetime, timing_distribution, timespan.
+    # a missing key is NOT an error.
+    assert "Found 3 errors" in result.output
```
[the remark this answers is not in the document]
badboy: Correct. 
I mean we _could_ add yet another schema rule that if BOTH are set we also fail, but ... I don't actually know how to "fail" then, hm.

PR #442, opened 2022-02-01
Bug 1751730 - Skip message if no metrics are found

badboy:
~~This can be for 2 reasons: no files passed or empty files passed.~~
~~Both cases are actually acceptable use of Glean:~~
~~You might want to use Glean _only_ for the baseline ping.~~
~~Especially initially, when you don't have metrics.~~

~~We still need to generate the build info so users can call~~
~~Glean.initialize with valid values.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Remove C# support ([#436](https://github.com/mozilla/glean_parser/pull/436)).
 - Add support for Rust code generation ([bug 1677434](https://bugzilla.mozilla.org/show_bug.cgi?id=1677434))
+- Report an error if no files are passed ([bug 1751730](https://bugzilla.mozilla.org/show_bug.cgi?id=1751730))
 
 ## 4.4.0
 
--- a/glean_parser/translate.py
+++ b/glean_parser/translate.py
@@ -136,6 +136,12 @@ def translate_metrics(
 
     input_filepaths = util.ensure_list(input_filepaths)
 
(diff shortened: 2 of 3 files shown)


glean_parser/lint.py
```diff
@@ -555,7 +555,9 @@ def glinter(
     errors += len([nit for nit in nits if nit.check_type == CheckType.error])
 
     if errors == 0:
-        print("✨ Your metrics are Glean! ✨", file=file)
+        if len(objs.value) > 0:
+            # Only report if we parsed some metrics.
+            print("✨ Your metrics are Glean! ✨", file=file)
```
Dexterp37, line 560: Should this fail if no file is provided?

badboy: I thought about it and put some explanation in the commit message:
No metrics fail is actually allowed behavior. If someone wants to use Glean just for the baseline ping they don't need custom metrics.

Dexterp37: Do we call glean_parser even though we have no file? I though we didn't even have the integration if the basic integration was all that was desired.

I agree that we should allow empty files.

badboy: Nevermind, it defaults to `false`

badboy: Which means I can depend on that. Let me update the PR

badboy:
Alright, @Dexterp37. What do you think about this?

Dexterp37 (approved):
Thanks!

PR #446, opened 2022-02-03
Bug 1753194 - add support for expiring metrics by major version

Dexterp37:
This introduces a new command line argument that allows the integrator to provide the major version number of the product.

If that's provided, glean_parse will only allow metrics to expire by major version instead than date.

@mdboom do you think we should bump the schema version? This feels a pretty safe change (famous last words), as we are increasing the set of allowed values instead of restricting more.

### DIFF
--- a/glean_parser/schemas/metrics.2-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.2-0-0.schema.yaml
@@ -270,12 +270,22 @@ definitions:
               metric expires. For example, `2019-03-13`. This date is checked at
               build time. Except in special cases, this form should be used so
               that the metric automatically "sunsets" after a period of time.
+            - `<major version>`: A positive integer representing the major
+              version the metric expires in. For example, `11`. The version is
+              checked at build time against the major provided to the
+              glean_parser and is only valid if a major version is provided at
+              built time. If no major version is provided at built time and
+              expiration by major version is used for a metric, an error is
+              raised.
             - `never`: This metric never expires.
             - `expired`: This metric is manually expired.
 
           The default may be overriden in certain applications by the
           `custom_validate_expires` and `custom_is_expired` configs.
-        type: string
+        oneOf:
+          - type: string
+          - type: integer
+            minimum: 0
 
       version:
         title: Metric version
--- a/glean_parser/schemas/metrics.2-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.2-0-0.schema.yaml
@@ -270,13 +270,15 @@ definitions:
               metric expires. For example, `2019-03-13`. This date is checked at
               build time. Except in special cases, this form should be used so
               that the metric automatically "sunsets" after a period of time.
-            - `<major version>`: A positive integer representing the major
-              version the metric expires in. For example, `11`. The version is
-              checked at build time against the major provided to the
+            - `<major version>`: An integer greater than 0 representing the
+              major version the metric expires in. For example, `11`. The
+              version is checked at build time against the major provided to the
               glean_parser and is only valid if a major version is provided at
-              built time. If no major version is provided at built time and
+              built time. If no major version is provided at build time and
               expiration by major version is used for a metric, an error is
               raised.
+              Note that mixing expiration by date and version is not allowed
+              within a product.
             - `never`: This metric never expires.
             - `expired`: This metric is manually expired.
 
@@ -285,7 +287,7 @@ definitions:
         oneOf:
           - type: string
           - type: integer
-            minimum: 0
+            minimum: 1
 
       version:
         title: Metric version
--- a/glean_parser/util.py
+++ b/glean_parser/util.py
@@ -374,7 +374,7 @@ def format_error(
         return f"{filepath}:\n{textwrap.indent(content, '    ')}"
 
 
-def parse_expires(expires: str) -> datetime.date:
+def parse_expiration_date(expires: str) -> datetime.date:
     """
     Parses the expired field date (yyyy-mm-dd) as a date.
     Raises a ValueError in case the string is not properly formatted.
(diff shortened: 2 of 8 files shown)


glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -270,12 +270,22 @@ definitions:
               metric expires. For example, `2019-03-13`. This date is checked at
               build time. Except in special cases, this form should be used so
               that the metric automatically "sunsets" after a period of time.
+            - `<major version>`: A positive integer representing the major
+              version the metric expires in. For example, `11`. The version is
+              checked at build time against the major provided to the
+              glean_parser and is only valid if a major version is provided at
+              built time. If no major version is provided at built time and
```
badboy, line 277:
```suggestion
              build time. If no major version is provided at build time and
```

glean_parser/util.py
```diff
@@ -374,21 +374,38 @@ def format_error(
...
-def is_expired(expires: str) -> bool:
+def parse_expiration_version(expires: str) -> int:
+    """
+    Parses the expired field version string as an integer.
+    Raises a ValueError in case the string does not contain a valid
+    positive integer.
+    """
+    try:
+        version_number = int(expires)
+        if version_number < 0:
```
badboy, line 399: Should we exclude `0` here? (`<= 0`)

badboy (approved):
Looking good. That took surprisingly few changes to make it work.

badboy:
Do you have follow-up bugs to bring that to the Gradle plugin and the iOS script?

badboy:
Retracting my r+ for the time being.
Only now that I saw the implementation and raised the above question I realized one tiny detail:
Focus and Firefox iOS use a version of `9000` for developer (and beta/testflight) builds.
They won't be able to start using expiry-by-version that way or it will just always break in dev builds.
We probably need to work with them to find out if they are eager to change that or if we need to find another way to do it.
(local dev builds at least always have `build_id=1`, so we could un-expire things if that's the case. That won't work for testflight builds though)

glean_parser/util.py
```diff
@@ -374,21 +374,38 @@ def format_error(
...
 
-def is_expired(expires: str) -> bool:
+def parse_expiration_version(expires: str) -> int:
+    """
+    Parses the expired field version string as an integer.
+    Raises a ValueError in case the string does not contain a valid
+    positive integer.
+    """
+    try:
+        version_number = int(expires)
```
mdboom, line 398: Do we want to implicitly convert from a string to an int here, or require that the versions be ints in the input file?  I can see this both ways, I guess, but the alternative would be `if isinstance(expires, int):`

glean_parser/util.py
```diff
@@ -397,20 +414,32 @@ def is_expired(expires: str) -> bool:
...
     """
     if expires in ("never", "expired"):
         return
 
-    date = parse_expires(expires)
+    if major_version is not None:
+        parse_expiration_version(expires)
+        # Don't need to keep parsing dates if expiration by version
+        # is enabled. We don't allow mixing dates and versions for a
+        # single product.
```
mdboom, line 439: I agree with this restriction.  Should we also document this in the description in the schema above (which ends up in the docs)?

tests/test_parser.py
```diff
@@ -613,6 +613,81 @@ def test_custom_expires():
...
+            },
+        )
+        list(all_metrics)
+
+    with pytest.raises(ValueError):
+        # Mixing expiration types must fail when expiring by date.
+        all_metrics = parser.parse_objects(ROOT / "data" / "mixed-expirations.yaml")
+        list(all_metrics)
+
+
```
mdboom, line 690: Can we add a test that checks (1) if no expiration version is specified, (2) specifying an expiration version on a metric fails?  I imagine it should be fine, since `99` isn't an iso date, but there might be some weird world in which that turns into `0099-01-01` or something.  Also, that case might throw an ISO parsing error to the user, and it would probably be better to throw an error that explains "can't use versions in the file without passing the --expire-by-version parameter` or something.

mdboom (changes requested):
Looks good.

To your question about schema version -- maybe we increment the minor part of semver, since it is still backward compatible?  It would give us am easier way to flag which `metrics.yaml` files have been migrated to the version-based approach.  (But don't have strong opinions about that -- there is additional complexity involved to changing it).

Dexterp37:
> Do you have follow-up bugs to bring that to the Gradle plugin and the iOS script?

Yes, that's happening as follow-up work on this bug (not the iOS part , for which I'll file a separate bug!).

glean_parser/util.py
Dexterp37: Good call.

glean_parser/util.py
Dexterp37: Good idea, yes, let me require ints in the file :)

tests/test_parser.py
Dexterp37: Added the test, we were already correctly raising the right error, something like `Invalid expiration date '99'. Must be of the form yyyy-mm-dd in UTC.`

Dexterp37:
> (But don't have strong opinions about that -- there is additional complexity involved to changing it).

This one last bit made me go the "easy path" and just not do any version bump :)

Dexterp37:
> Retracting my r+ for the time being. Only now that I saw the implementation and raised the above question I realized one tiny detail: Focus and Firefox iOS use a version of `9000` for developer (and beta/testflight) builds. They won't be able to start using expiry-by-version that way or it will just always break in dev builds. We probably need to work with them to find out if they are eager to change that or if we need to find another way to do it. (local dev builds at least always have `build_id=1`, so we could un-expire things if that's the case. That won't work for testflight builds though)

Good catch! I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1753975) and noted your comment. As discussed over Matrix, let's move on with this implementation and figure out the iOS part by talking to the devs :)

badboy (approved):

Dexterp37:
Thanks for the review! I'll go ahead and merge, then cut a release.

PR #450, opened 2022-02-08
Fix the expiration by version logic

Dexterp37:

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -643,6 +643,10 @@ def test_expire_by_major_version():
                     "type": "boolean",
                     "expires": 7,
                 },
+                "metric_expired_edge": {
+                    "type": "boolean",
+                    "expires": 15,
+                },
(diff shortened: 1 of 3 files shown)


tests/test_parser.py
```diff
@@ -641,15 +641,15 @@ def test_expire_by_major_version():
...
+                    "expires": 7,
                 },
                 "metric_expired": {
                     "type": "boolean",
                     "expires": "expired",
                 },
                 "metric": {
                     "type": "boolean",
-                    "expires": 7,
+                    "expires": 18,
```
badboy, line 652: Can you add one more metric that expires in v15?
Just so we have a test that's explicitly the edge case ("expires" defines the first version in which this metric is expired)

badboy (approved):

tests/test_parser.py
Dexterp37: Sure, done!

PR #461, opened 2022-03-14
Mark the Glean team as author of this package

badboy:
Wasn't sure about Development Lead.
It for sure includes me, but also you or Bea?


Dexterp37 (approved):
So long!

PR #469, opened 2022-03-30
Bug 1761791 - Fix issues with Swift template

brizental:
See commits for list of issues fixed.

### Pull Request checklist ###

- [ ] **Quality**: This PR builds and tests run cleanly
  - `make test` runs without emitting any warnings
  - `make lint` runs without emitting any errors
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes a changelog entry to `CHANGELOG.md` or an explanation of why it does not need one
  - Any breaking changes to language binding APIs are noted explicitly

PR #470, opened 2022-04-01
Revert "Bug 1761791 - Fix issues with Swift template"

brizental:
Reverts mozilla/glean_parser#469

These changes ended up being unnecessary. They change the way to access pings and break the template when pings have reasons. Other workarounds can be used to access metrics that are module private, such as having a different Metrics.swift file per module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
+- BUGFIX: Revert changes made on v5.1.1. 
+    - The issues addressed by those changes, were non-issues and result of misuse of the APIs.
+
 ## 5.1.1
+
+- BUGFIX: Fix issues with Swift templates ([bug 1749494](https://bugzilla.mozilla.org/show_bug.cgi?id=1749494))
+    - Make metrics and pings all `public`
+    - Make pings `static`
 
 ## 5.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## 5.1.1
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,10 +2,6 @@
 
 ## 5.1.1
 
-- BUGFIX: Fix issues with Swift templates ([bug 1749494](https://bugzilla.mozilla.org/show_bug.cgi?id=1749494))
```
Dexterp37, line 5: Instead of removing something from a release that was already cut, we should add a new changelog entry explaining that 5.1.1 was broken for Swift and explaining what was reverted.

brizental: Hehe, oops I just clicked the revert button. Nice catch!

Dexterp37 (approved):
r+wc

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+- BUGFIX: Revert changes made on v5.1.1. 
+    - The issues addressed by those changes, were non-issues and result of misuse of the APIs.
```
Dexterp37, line 4: Is there anything we need to clarify in the glean book?

brizental: I will check that later today. Might be worth noting that the metrics are module private, although that miht be obvioust o a more experienced Swift programmer.

travis79: The `internal` access is indeed "module internal/private", and might be worth calling out. Had I been thinking about that and how Firefox iOS was structured, we wouldn't have gone this route. I would also be glad to see more documentation/comments to help guide us the next time something like this comes up. Feel free to file a bug around adding some clarification to using Glean in a project that has multiple targets/modules in Xcode, definitely a little more challenging getting everything just right than I expected.

PR #473, opened 2022-04-13
[Bug 1749471] List options accepted by each translation target in glean_parser

perrymcmanis144:
@badboy ~tests seem to be failing locally for me; but they arent in any files I've changed~ chmod to the rescue

also the way I handled this is that if you use the --list-options flag ala

`glean_parser translate --list-options`

 it will show a paged view of them and then exit. This has the nicety of avoiding having to set a language target (which is normally required), however this is just how it makes sense to me. I could change this implementation to work as

`glean_parser translate -o output_loc/ -f lang --list-options`

it would show you only the options for that language.  Im open to suggestion as to which is more ergonomic.

### DIFF
--- a/glean_parser/__main__.py
+++ b/glean_parser/__main__.py
@@ -24,7 +24,10 @@ from . import validate_ping
 from . import translation_options
 
 
-@click.command()
+CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
+
+
+@click.command(context_settings=CONTEXT_SETTINGS)
 @click.argument(
     "input",
     type=click.Path(exists=False, dir_okay=False, file_okay=True, readable=True),
@@ -46,10 +49,13 @@ from . import translation_options
 @click.option(
     "--option",
     "-s",
-    help="backend-specific option. Must be of the form key=value",
+    help="Backend-specific option. Must be of the form key=value.\
+ Pass 'help' for valid options",
     type=str,
     multiple=True,
     required=False,
+    is_eager=True,
+    callback=translation_options.translate_options,
 )
 @click.option(
     "--allow-reserved",
@@ -75,14 +81,6 @@ from . import translation_options
     type=click.INT,
     required=False,
 )
-@click.option(
-    "--list-options",
-    is_flag=True,
-    default=False,
-    help="List valid options for target language",
-    is_eager=True,
-    callback=translation_options.translate_options,
-)
 def translate(
     input,
     format,
@@ -92,7 +90,6 @@ def translate(
     allow_missing_files,
     require_tags,
     expire_by_version,
-    list_options,
 ):
     """
     Translate metrics.yaml and pings.yaml files to other formats.
--- a/glean_parser/__main__.py
+++ b/glean_parser/__main__.py
@@ -21,6 +21,7 @@ from . import data_review as mod_data_review
 from . import lint
 from . import translate as mod_translate
 from . import validate_ping
+from . import translation_options
 
 
 @click.command()
@@ -73,6 +74,13 @@ from . import validate_ping
     help="Expire metrics by version, with the provided major version.",
     type=click.INT,
     required=False,
+)
+@click.option(
+    "--list-options",
+    is_flag=True,
+    help="List valid options for target language",
+    is_eager=True,
+    callback=translation_options.translate_options,
 )
 def translate(
     input,
--- a/glean_parser/__main__.py
+++ b/glean_parser/__main__.py
@@ -78,6 +78,7 @@ from . import translation_options
 @click.option(
     "--list-options",
     is_flag=True,
+    default=False,
     help="List valid options for target language",
     is_eager=True,
     callback=translation_options.translate_options,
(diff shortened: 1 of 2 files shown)


glean_parser/__main__.py
```diff
@@ -74,6 +75,14 @@
     type=click.INT,
     required=False,
 )
+@click.option(
+    "--list-options",
```
badboy, line 79: Any chance we can just make this part of the `--help` output? Having yet another option for this seems unexpected.

perrymcmanis144: for when you get back, I did some investigation and here's what I worked out:

- I _can_ hijack the help, but it's not so trivial. I was having loads of issue actually accessing the commands & even if I could I would have to programmatically build it but not so difficult. Kind of a time sink working out how to get the cli commands for the subcommand.
- In general help seems really hidden; it would be really nice if I could just chuck the extra text at the end. This is one of those cases where pip versus shipping a package is a real bummer. probably solvable with time
- I was able to enable -h for help easily
- I have removed --list-options & instead used that pattern such that if you type this command you will get the view in hopes this represents an ergonomic improvement over having two options

`glean_parser translate --options help`

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
+import pydoc
+
+
+def translate_options(ctx, param, value):
+    text = """Target language options for Translate function
+
+These are backend specific and optional, provide as key:value
+
+Rust:
+- not used
```
badboy, line 10:
```suggestion
Rust: no options.
```

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
...
+Swift:
```
badboy, line 11: Let's add a bit of spacing:

```suggestion

Swift:
```

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
...
+- namespace: The namespace to generate metrics in
```
badboy, line 12: Further below you wrap names in backticks. Let's make it consistent.

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
...
+- glean_namespace: The namespace to import Glean from
+- allow_reserved: When True, this is a Glean-internal build
+- with_buildinfo: If "true" the `GleanBuildInfo` is generated.
+        Otherwise generation of that file is skipped.
+        Defaults to "true".
+- build_date: If set to `0` a static unix epoch time will be used.
+        If set to a ISO8601 datetime string (e.g. `2022-01-03T17:30:00`)
+        it will use that date.
+        Other values will throw an error.
+        If not set it will use the current date & time.
```
badboy, line 22:
```suggestion
- `namespace`: The namespace to generate metrics in
- `glean_namespace`: The namespace to import Glean from
- `allow_reserved`: When True, this is a Glean-internal build
- `with_buildinfo`: If "true" the `GleanBuildInfo` is generated.
        Otherwise generation of that file is skipped.
        Defaults to "true".
- `build_date`: If set to `0` a static unix epoch time will be used.
        If set to a ISO8601 datetime string (e.g. `2022-01-03T17:30:00`)
        it will use that date.
        Other values will throw an error.
        If not set it will use the current date & time.
```

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
...
+Kotlin:
```
badboy, line 23:
```suggestion

Kotlin:
```

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
...
+- `namespace`: The package namespace to declare at the top of the
+        generated files. Defaults to `GleanMetrics`.
+- `glean_namespace`: The package namespace of the glean library itself.
+        This is where glean objects will be imported from in the generated
+        code.
+JavaScript:
```
badboy, line 29:
```suggestion

JavaScript:
```

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
...
+        This option is mandatory when targeting Qt. Note that the version
+        string must only contain the major and minor version i.e. 0.14.
+- `with_buildinfo`: If "true" a `gleanBuildInfo.(js|ts)` file is generated.
+        Otherwise generation of that file is skipped. Defaults to "false".
+- `build_date`: If set to `0` a static unix epoch time will be used.
+        If set to a ISO8601 datetime string (e.g. `2022-01-03T17:30:00`)
+        it will use that date.
+        Other values will throw an error.
+        If not set it will use the current date & time.
+Markdown:
```
badboy, line 42:
```suggestion

Markdown:
```

glean_parser/translation_options.py
```diff
@@ -0,0 +1,51 @@
...
+- `project_title`: The projects title.
```
badboy, line 43:
```suggestion
- `project_title`: The project's title.
```

badboy (changes requested):
I like this option! A bit of formatting, then we can land this.

badboy (approved):

PR #495, opened 2022-06-09
Some fixes for Rust code generation

travis79:
- Changes the Rust jinja template to use the `extra_metric_args` format more like the other language templates.
- Adds full qualification of `String` to `std::string::String` to silence Rust errors


glean_parser/templates/rust.jinja2
```diff
@@ -32,10 +32,10 @@ pub struct {{ obj.name|Camelize }}{{ suffix }} {
 impl ExtraKeys for {{ obj.name|Camelize }}{{ suffix }} {
     const ALLOWED_KEYS: &'static [&'static str] = {{ obj.allowed_extra_keys|extra_keys }};
 
-    fn into_ffi_extra(self) -> ::std::collections::HashMap<String, String> {
+    fn into_ffi_extra(self) -> ::std::collections::HashMap<std::string::String, std::string::String> {
```
badboy, line 35: Let's use full qualificaation here, as the hashmap also uses with a `::` prefix

```suggestion
    fn into_ffi_extra(self) -> ::std::collections::HashMap<::std::string::String, ::std::string::String> {
```

badboy: I think that should work at least.

PR #523, opened 2022-10-17
Bug 1793777 - Support rate metric for Glean.js/JavaScript

rosahbruno:
When trying to upgrade the version of `glean_parser` that Glean.js is using, we were seeing errors that the `rate` metric was not supported. The `rate` metric already exists in Glean.js and works in previous versions, but the newer versions of `glean_parser` did not support the `rate` metric for JS.

This PR
- Updates the javascript parser to correctly serialize the `rate` metric
- Removes the code that was throwing `rate` errors and skipping `rate` metrics for Glean.js (https://github.com/mozilla/glean_parser/commit/19a14a7604aaae9e8cfca83daed9e021ce10ec70)

### Testing
Testing Glean.js with different versions of `glean_parser`
- 5.1.0 (current version used by glean.js) - rate works fine
- 6.2.0 (latest version) - tests fail, error saying `rate` metric is unsupported
- this branch - rate works fine, no errors again

I tested this by
- pushing this code up to this branch
- creating a `venv` locally and using pip to install `glean_parser` directly from this branch (`python3 -m pip install git+ssh://git@github.com/rosahbruno/glean_parser@1793777-JS-rate-metric#egg=glean_parser`)

### DIFF
--- a/glean_parser/javascript.py
+++ b/glean_parser/javascript.py
@@ -148,16 +148,11 @@ def output(
                         it will use that date.
                         Other values will throw an error.
                         If not set it will use the current date & time.
-        - `fail_rates`: When `true` it fails when encountering rate metrics.
-                        When `false` it will warn and skip rate metrics.
-                        Defaults to "true".
     """
 
     if options is None:
         options = {}
 
-    fail_rates = options.get("fail_rates", "true").lower() == "true"
-    fail_rate_level = "ERROR" if fail_rates else "WARN"
     platform = options.get("platform", "webext")
     accepted_platforms = ["qt", "webext", "node"]
     if platform not in accepted_platforms:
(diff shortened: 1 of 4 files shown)


glean_parser/javascript.py
```diff
@@ -37,6 +51,17 @@ def iterencode(self, value):
                     yield from self.iterencode(subvalue)
                     first = False
                 yield "]"
+            elif isinstance(value, metrics.Rate):
```
badboy, line 54: Why does rate require special handling?

badboy (commented):
one question, but otherwise go ahead.

badboy (approved):

glean_parser/javascript.py
rosahbruno: The first 2 `rate` metrics work fine, but since `rate_external` has `denominator_metric` it cannot serialize that property (https://github.com/mozilla/glean_parser/blob/main/tests/data/all_metrics.yaml#L113).

It was throwing an error - `TypeError: Object of type Rate is not JSON serializable`. The difference is that in most cases it is serializing the properties of all the objects, which are serializable types. Since `rate_external` has a `rate` metric are one of its properties, rather than trying to serialize a string or number, it now has to try and serialize another `rate` object, which it can't by default.

For reference, this block is the same as what we are adding for both `swift` and `kotlin` to make sure we serialize properly
- swift - https://github.com/mozilla/glean_parser/blob/main/glean_parser/swift.py#L74
- kotlin - https://github.com/mozilla/glean_parser/blob/main/glean_parser/kotlin.py#L72

PR #549, opened 2023-01-23
Bug 1811857 - Fully remove support for the old events API

badboy:
For https://github.com/mozilla/glean/pull/2346
Based on #551 

---

Can't just change it in the schema because that would be breaking.
I still need to check if this breaks anything in probe-scraper. I guess it would break because we can't parse old files anymore?
In that case we would need to bump the schema to 3-0-0 and migrate all users eventually.

### DIFF
--- a/tests/data/events_with_types.yaml
+++ b/tests/data/events_with_types.yaml
@@ -5,23 +5,6 @@
 $schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
 
 core:
-  event_example:
-    type: event
-    description: >
-      Just testing events
-    bugs:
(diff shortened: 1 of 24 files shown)


tests/data/events_with_types.yaml
```diff
@@ -18,8 +18,10 @@ core:
     extra_keys:
       key1:
         description: "This is key one"
+        type: string
```
travis79, line 21: I'm not seeing where this metric is used in any of the tests, only the preference_toggled one seems to be checked by `assert`s

travis79 (approved):

PR #560, opened 2023-03-06
Bug 1820648 - Sort generated Metrics, Pings and Tags

perrymcmanis144:
Here's a take on making sure we have a deterministic order from `glean_parser`.

Ordered Dict is a weird one. It preserves insertion order, true, but in doing some research it's had rather a lot of changes recently and performance hasn't always been great. While this isn't a perf critical application, this method (calling sort on the `.items()` iterator) is stable, runs once each for Metrics/Pings and Tags, and will always result in a consistent order (alphabetical).

Interestingly, regular `dict` ordering is _supposed_ to be guaranteed as of 3.6, except that it's unsettled what that order should be. 

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -844,6 +844,51 @@ def test_rates():
     )  # Hasn't been transformed to "denominator" yet
 
 
+def test_ping_ordering():
+    contents = parser.parse_objects(
+        [ROOT / "data" / "pings.yaml"],
+        config={"allow_reserved": False},
+    )
+
+    errs = list(contents)
+    assert len(errs) == 0
+
+    pings = list(contents.value["pings"].keys())
+
+    assert pings == sorted(pings)
+
+
+def test_metric_ordering():
+    all_metrics = parser.parse_objects(
+        [ROOT / "data" / "ordering.yaml"], config={"allow_reserved": False}
+    )
+
+    errs = list(all_metrics)
+    assert len(errs) == 0
+
+    category = all_metrics.value["testing.ordering"]
+
+    assert len(category.values()) > 0
+    test_metrics = [f"{m.category}.{m.name}" for m in category.values()]
+
+    # Alphabetically ordered
+    assert test_metrics == [
+        "testing.ordering.a_second_test_metric",
+        "testing.ordering.first_test_metric",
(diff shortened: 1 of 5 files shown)


chutten (changes requested):
Needs CHANGELOG, and could probably benefit from a dedicated test instead of the accidental one from ratio : )

perrymcmanis144:
@chutten and now there is a test (both correct and incorrect order) at least for some metrics ordering since you're ok with this strategy.

it's my EOD, if you'd like me to add additional tests to cover ping & tag ordering tomorrow let me know, I assume they are worth having but if you think this is adequate then that would wfm too.

perrymcmanis144:
Ready for proper review. Tests in for all 3 parts of the parser that will be ordered.

tests/test_parser.py
```diff
@@ -844,6 +844,58 @@ def test_rates():
...
+        [ROOT / "data" / "ordering.yaml"], config={"allow_reserved": False}
+    )
+
+    errs = list(all_metrics)
+    assert len(errs) == 0
+
+    category = all_metrics.value["testing.ordering"]
+
+    assert len(category.values()) > 0
+    test_metrics = [f"{m.category}.{m.name}" for m in category.values()]
```
chutten, line 872: Uh, but won't these already be sorted on input, since they're in order in the yaml? Maybe we need a `fourth_test_metric` which is lexicographically second?

tests/test_parser.py
```diff
@@ -844,6 +844,58 @@ def test_rates():
...
+        "testing.ordering.second_test_metric",
+        "testing.ordering.third_test_metric",
+    ]
+
+    # Switch the ordering such that the metrics are not alphabetical
+    assert test_metrics != [
+        "testing.ordering.second_test_metric",
+        "testing.ordering.a_test_metric",
+        "testing.ordering.third_test_metric",
+    ]
```
chutten, line 886: I'm confused by this. Of course if it's equal to the first list it can't be equal to a second, differently-ordered list?

chutten:
Also, you should probably rename this PR to point at your glean_parser-specific bug instead of the EventGIFFTMap one.  The bot attached this PR to it, so folks might get a little confused. (You can mark the attachment obsolete in its Details)

PR #565, opened 2023-03-20
Bug 1816526 - Remove record_event_by_id_with_strings

perrymcmanis144:
This function is no longer being used downstream so removing it from the template

@chutten I don't know if this was something that was ever expected to be used beyond FOG, so my instinct is that this one does not deserve a changelog entry, but I'm happy to add one if you think it does. 


chutten (dismissed):
This would never be used by FOG, as it has its own rust template. Instead, this could be used by non-FOG Rust consumers (like VPN). You might want to ask @brizental and you'll definitely want to add a `CHANGELOG` for it. I'm not sure if this will require a major version bump on glean_parser, but semver might recommend it.

badboy:
VPN has its own templates (just like FOG), using its own version of that function: https://github.com/mozilla-mobile/mozilla-vpn-client/blob/f86e748168964bb8ab68eda6c942efcf0012bac4/qtglean/glean_parser_ext/templates/rust.jinja2#L114

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - BREAKING CHANGE: Remove exposed `lint_yaml_files` function ([#580](https://github.com/mozilla/glean_parser/pull/580))
+- Rust: Removed \_\_glean\_metric\_maps from the Rust Jinja template. This functionality is better placed downstream ([Bug 1816526](https://bugzilla.mozilla.org/show_bug.cgi?id=1816526))
```
badboy, line 6:
```suggestion
- Rust: Removed `__glean_metric_maps` from the Rust Jinja template. This functionality is better placed downstream ([Bug 1816526](https://bugzilla.mozilla.org/show_bug.cgi?id=1816526))
```

badboy (approved):

PR #587, opened 2023-06-05
Bug 1839640 - Add new metric type: object

badboy:
This is the most bare-bones version for now.
It validates the `structure` and then generates the corresponding Rust code.

### DIFF
--- a/tests/data/object.yaml
+++ b/tests/data/object.yaml
@@ -1,0 +1,28 @@
+# Any copyright is dedicated to the Public Domain.
+# https://creativecommons.org/publicdomain/zero/1.0/
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+activity.stream:
+  tiles:
+    type: object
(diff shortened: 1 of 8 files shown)


tests/data/object.yaml
```diff
@@ -0,0 +1,96 @@
...
+            properties:
+              count:
+                type: number
+              string:
+                type: string
+
+crash.stack:
+  threads:
+    type: object
+    description: All threads' frame information.
```
chutten, line 73:
```suggestion
    description: |
      All threads' frame information.
      e.g. `[{frames: [{module_index: 0, ip: "0xdecafc0ffe", trust: "scan"}]}]`
```

chutten (changes requested):
Is the test that the yaml is parsed at all? Is there a test of the rust codegen?

travis79 (approved):
Looks good to me, I guess we will add support for other languages as needed?

badboy:
> Looks good to me, I guess we will add support for other languages as needed?

yes. I'll file bugs.

PR #595, opened 2023-07-14
FXA-7922 - Add server-side JavaScript outputter

akkomar:
Proposal: https://docs.google.com/document/d/1TV7FXh1Sfxe4I4LdlOpiGIQp1hXwWO_n9ezG86Hyz5Y/edit#heading=h.5x0d5h95i329


glean_parser/javascript_server.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        return
+
+    # go through all metrics in objs and build a map of
+    # ping->list of metric categories->list of metrics
+    ping_to_metrics: Dict[str, Dict[str, List[metrics.Metric]]] = defaultdict(dict)
+    for _category_key, category_val in objs.items():
+        for _metric_name, metric in category_val.items():
+            if isinstance(metric, metrics.Metric):
+                if metric.type not in ["string"]:
```
badboy, line 91: It might make sense to pull out this list into a constant for easier modification later. And this would need some docs on when and how we can extend that

glean_parser/templates/javascript_server.jinja2
```diff
@@ -0,0 +1,120 @@
...
+        {% endfor %}
+      },
+      ping_info: {
+        seq: 0, // this is required, however doesn't seem to be useful in server context
+        start_time: timestamp,
+        end_time: timestamp,
+      },
+      // `Unknown` fields below are required in the Glean schema, however they are not useful in server context
+      client_info: {
+        telemetry_sdk_build: 'glean_parser v{{ parser_version }}',
```
badboy, line 73: Feels a bit weird to use the glean_parser here as the SDK version, but it's also not ... wrong. Hmmm

glean_parser/templates/javascript_server.jinja2
```diff
@@ -0,0 +1,120 @@
...
+    let ping = {
+      document_namespace: this._applicationId,
+      document_type: '{{ ping }}',
+      document_version: '1',
+      document_id: uuidv4(),
+      payload: eventPayload,
+    };
+
+    // this is similar to how FxA currently logs with mozlog: https://github.com/mozilla/fxa/blob/4c5c702a7fcbf6f8c6b1f175e9172cdd21471eac/packages/fxa-auth-server/lib/log.js#L289
+    this._log.info(GLEAN_EVENT_MOZLOG_TYPE, ping);
```
badboy, line 95: How does this then get to our pipeline?

badboy (commented):
Honestly I kinda expected the parser part to only generate some code that calls into a module somewhere on the outside for assembling the payload.
But looking at this it's pretty much self-contained, so it's ok IMO.

I could see this living as its own tool, using glean_parser as a dependency, but that would require additional work to release it, consume it, etc.
So again, because it's self-contained, I think it's ok to keep it here.

glean_parser/templates/javascript_server.jinja2
badboy: Asking because the proposal discussed significantly easier generated code that doesn't strictly adhere to our json ping schema. there is was specified that some "GCP Cloud logging to bigquery" adapter would be used, meaning this never goes through the early parts of the pipeline.

glean_parser/javascript_server.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # In this environment we don't use a concept of measurement window for collecting
+    # metrics. Only "events as pings" are supported.
+    # For each ping we generate code which contains all the logic for assembling it
+    # with metrics, serializing, and submitting. Therefore we don't generate classes for
+    # each metric as in standard outputters.
+    if "pings" not in objs:
+        print(
+            "❌ No ping definition found. Server-side environment is simplified and this"
+            + "parser doesn't generate individual metric files."
```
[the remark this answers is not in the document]
akkomar: Good point, I'll expand the error message to mention this.

glean_parser/templates/javascript_server.jinja2
akkomar: These logs will be delivered to PubSub via Cloud Logging Sink, processed by the modified Decoder ([PR opened](https://github.com/mozilla/gcp-ingestion/pull/2400)) which will deliver them to `structured-decoded` topic. Next steps are standard pipeline processing (BQ live tables -> stable tables).

badboy: Thanks, that was the missing context.

glean_parser/templates/javascript_server.jinja2
badboy: Yeah, after talking with Travis the other day I'm not blocking on this really. It's correct; and if in the future we determine this should report something different or live elsewhere everything here is self-contained enough to pull it off easily

glean_parser/templates/javascript_server.jinja2
```diff
@@ -0,0 +1,154 @@
...
+{% if lang == "typescript" %}
+// @types/uuid and mozlog types definitions are required in devDependencies
+// for the latter see https://github.com/mozilla/fxa/blob/85bda71cda376c417b8c850ba82aa14252208c3c/types/mozlog/index.d.ts
+{% endif %}
+import { v4 as uuidv4 } from 'uuid';
+import mozlog{% if lang == "typescript" %}, { Logger }{% endif %} from 'mozlog';
+
+const GLEAN_EVENT_MOZLOG_TYPE = 'glean-server-event';
+
+let _logger{% if lang == "typescript" %}: Logger{% endif %};
```
badboy, line 19: The first call to any factory method will set this and thus prescribe which `logger_options` are used.
If there's multiple pings and a user calls the factory method with different `logger_options` the order of calls influences what options are set.

I assume in the current usage this wouldn't be an issue because (a) there's only a single ping, (b) even if it were multiple pings it would be the same logger options, is that correct?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - BREAKING CHANGE: Remove exposed `lint_yaml_files` function ([#580](https://github.com/mozilla/glean_parser/pull/580))
 - Rust: Removed `__glean_metric_maps` from the Rust Jinja template. This functionality is better placed downstream ([Bug 1816526](https://bugzilla.mozilla.org/show_bug.cgi?id=1816526))
 - New lint: check that all referenced pings are known ([#584](https://github.com/mozilla/glean_parser/pull/584))
+- Add basic server-side JavaScript outputter ([FXA-7922](https://mozilla-hub.atlassian.net/browse/FXA-7922))
```
badboy, line 8:
```suggestion
- Add experimental server-side JavaScript outputter ([FXA-7922](https://mozilla-hub.atlassian.net/browse/FXA-7922))
```

Let's call it experimental.

tests/data/fxa-server-metrics.yaml
```diff
@@ -0,0 +1,21 @@
+# Any copyright is dedicated to the Public Domain.
+# https://creativecommons.org/publicdomain/zero/1.0/
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+event:
```
badboy, line 7: Ooooooh, I was a bit thrown off by this naming. It's actually a string metric inside.

badboy (changes requested):
Minimal last nit and a question.

glean_parser/templates/javascript_server.jinja2
akkomar: Yes, I'm aware this approach is not ideal. At the moment we don't expect to have multiple pings, but even in such case logger options should remain constant.

tests/data/fxa-server-metrics.yaml
akkomar: @badboy would `fxa-server-string-metric.yaml` be preferred?

badboy: Nah, it's fine this way.

badboy (approved):

PR #598, opened 2023-07-28
Update server-side js/ts output template

chenba:
* Replace a TS `any` type with a new type
* Replace a few instances of `let` (that some linters might complain about) with `const`

    - I see warnings related to Python dependencies but nothing related to my changes
  - As far as I can tell there are no tests for the template file I updated
    - I don't think the small edits in this PR need a changelog entry

### DIFF
--- a/glean_parser/templates/javascript_server.jinja2
+++ b/glean_parser/templates/javascript_server.jinja2
@@ -15,6 +15,9 @@ import { v4 as uuidv4 } from 'uuid';
 import mozlog{% if lang == "typescript" %}, { Logger }{% endif %} from 'mozlog';
 
 const GLEAN_EVENT_MOZLOG_TYPE = 'glean-server-event';
+{% if lang == "typescript" %}
+type LoggerOptions = { app: string; fmt?: 'heka' };
+{% endif %}
 
 let _logger{% if lang == "typescript" %}: Logger{% endif %};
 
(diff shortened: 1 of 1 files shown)


chenba:
@akkomar what do you think?

glean_parser/templates/javascript_server.jinja2
```diff
@@ -15,6 +15,9 @@ import { v4 as uuidv4 } from 'uuid';
 import mozlog{% if lang == "typescript" %}, { Logger }{% endif %} from 'mozlog';
 
 const GLEAN_EVENT_MOZLOG_TYPE = 'glean-server-event';
+{% if lang == "typescript" %}
+type LoggerOptions = { app: string; fmt?: 'heka' };
```
akkomar, line 19:
```suggestion
type LoggerOptions = { app: string };
```
I'd remove format from here since ingestion pipeline expects heka and this field is overridden in line 53/56 below.

akkomar (commented):
LGTM but check my comment about `fmt` - not sure if exposing it to change the format would be useful during development?

glean_parser/templates/javascript_server.jinja2
chenba: `fmt` can be either undefined or the literal string 'heka', which is how it's used at the moment.  I guess it is not particularly helpful to the user/caller since it is always overridden as you mentioned.  But the problem is that the type checker doesn't like that `fmt` is used in the assignment without it being declared.

badboy (approved):

PR #617, opened 2023-08-16
Bug 1848902 - Drop MarkupSafe dependency version restriction

badboy:
We added that a while ago because MarkupSafe 2.1 dropped Python 3.6 support and m-c required <2.
Therefore this restriction gaves us a range of versions to work with.

Since then we dropped Python 3.6 support anyway. And hopefully this should work in m-c too.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,6 @@ requirements = [
     "Click>=7",
     "diskcache>=4",
     "Jinja2>=2.10.1",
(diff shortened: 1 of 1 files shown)


badboy:
A naive "copy this checkout into m-c" (as per [docs](https://firefox-source-docs.mozilla.org/toolkit/components/glean/dev/updating_parser.html)) seems to work. Let's go with this.

travis79 (approved):

PR #620, opened 2023-08-18
feat(ruby): adding elements for ruby server events

braunreyes:

Resolves: [SOCIALPLAT-477](https://mozilla-hub.atlassian.net/browse/SOCIALPLAT-477)

In order to keep the special Mastodon logging logic in-sync with the latest Glean functionality, we need to create a translation process to auto-create the proper Ruby code.  This will be based off the javascript server version and currently only support ‘string’ metric types on for this initial version.


tests/test_ruby_server.py
```diff
@@ -0,0 +1,193 @@
...
+        "ruby_server",
+        tmpdir,
+    )
+
+    assert set(x.name for x in tmpdir.iterdir()) == set(["server_events.rb"])
+
+    # Make sure string metric made it in
+    with (tmpdir / "server_events.rb").open("r", encoding="utf-8") as fd:
+        content = fd.read()
+        content_1 = """# This Source Code Form is subject to the terms of the Mozilla Public
```
badboy, line 76: An inline full-generate code file will be awful to maintain. Probably best if we put that in an external file that the test reads in and add some code to update that file.

tests/test_ruby_server.py
```diff
@@ -0,0 +1,193 @@
...
+# end"""
+        assert content_1.strip().replace(" ", "").replace("\n", "").replace(
+            "\\n", ""
+        ) in content.strip().replace(" ", "").replace("\n", "").replace("\\n", "")
+        assert content_2.strip().replace(" ", "").replace("\n", "").replace(
+            "\\n", ""
+        ) in content.strip().replace(" ", "").replace("\n", "").replace("\\n", "")
+        assert content_3.strip().replace(" ", "").replace("\n", "").replace(
+            "\\n", ""
+        ) in content.strip().replace(" ", "").replace("\n", "").replace("\\n", "")
```
badboy, line 193: That's a whole lot of whitespace replacing and it's completely unclear why that is happening. Why can't we compare the content exactly?

braunreyes: this is just an artifact of splitting up the comparison text inline.  Lets see how the file comparison works.

tests/test_ruby_server.py
braunreyes: implemented

braunreyes:
data review created: https://bugzilla.mozilla.org/show_bug.cgi?id=1851692

braunreyes:
@badboy can we revisit this review?

Dexterp37:
@braunreyes hey! Can I get some more background on this? How does this fit into the other efforts for using Glean on the social platform? Can we jump on a meeting?

braunreyes:
@badboy added an initial pass at trying to implement server side with the event stream pattern

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,138 @@
...
+    when /solaris/
+      'Solaris'
+    when /bsd/
+      'BSD'
+    else
+      RbConfig::CONFIG['host_os']
+    end
+  end
+  def self.version
+    version = RbConfig::CONFIG['build_os'].match('\d{1,2}\.*\d{0,2}\.*\d{0,3}').to_s
```
badboy, line 40: Is the version guaranteed to be in that format for all OSs?

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,138 @@
...
+    if !version.nil?
```
badboy, line 41: Haven't done proper Ruby in a while, but shouldn't this be the same:

```suggestion
    if version
```

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,138 @@
...
+        version.to_s
```
badboy, line 42: `.to_s` is already used in the match line above, so this is a no-op
```suggestion
        version
```

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,138 @@
...
+  {% for event_type, event_object in event_types.items() %}
+  class {{ (event_category + '.' + event_type) | event_class_name }}
+    # see comments for init parameter descriptions
+    def initialize(application_id:, app_display_version:, app_channel:, user_agent:, ip_address:, event_extra)
+      @application_id = application_id # string - Application Id to identify application per Glean standards
+      @app_display_version = app_display_version # string - Application Id to identify application per Glean standards
+      @app_channel = app_channel # string - Application Id to identify application per Glean standards
+      @user_agent = user_agent # string - The user agent per Glean standards
+      @ip_address = ip_address # string - The IP address that will be used to decode Geo information and scrubbed at ingestion per Glean standards
+      @event_extra = event_extra # string - JSON blob containing the backend event payload
```
badboy, line 61: Why are we passing `event_extra` into the constructor?

actually let me pull this out into a full comment aside from the code review.

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,138 @@
...
+        'document_namespace' => @application_id,
+        'document_type' => 'events',
+        'document_version' => '1',
+        'document_id' => SecureRandom.uuid,
+        'user_agent' => @user_agent,
+        'ip_address' => @ip_address,
+        'payload' => serialized_event_payload,
+      }
+      # This structure is similar to how FxA currently logs with mozlog: https://github.com/mozilla/fxa/blob/4c5c702a7fcbf6f8c6b1f175e9172cdd21471eac/packages/fxa-auth-server/lib/log.js#L289
+      logger = Logger.new($stdout)
```
badboy, line 100: Is logging to `$stdout` really what we want? How are applications supposed to configure this?

badboy (changes requested):
Before I jump back into the full code review I think we should work back a little:

What is the Ruby API that we want to get to?
How is it supposed to be used?

The javascript_backend API is supposed to be called something like this:

```js
const { createMastodonActionEvent } = require("generated/glean.js");

let glean = createMastodonActionEvent({
  applicationId: "my-app",
  logger_options: { /* ... */ }
});
// `glean` should be stored by the application for later use,
// e.g. in some server context or global.

// later:
glean.record({
  some_metric: its_value,
  /* more data ... */
});
```

Note that the javascript_backend currently only supports `string` metrics.
It creates one class per ping.

In general the API that Glean would generate is an object per event and the `record` function takes in the event extras.

High-level things like the application ID, display version and the logger configuration probably should only be done once, so I could image something like:

```ruby
class GleanLog
  def initialize(app_id, display_version, logger_options)
    @app_id = app_id
    @display_version = display_version
    @logger = Logger.new(logger_options)
 
    # Generated events
    @mastodon_action = MastodonAction.new(self)
  end
  
  def _record(event)
    # ... snip
    # do all the logic to construct the payload and log it
  end

  attr_accessor :mastodon_action
end

class MastodonAction
  def initialize(glean)
    @glean = glean
  end

  def record(path:, account_id:, status_code:)
    # ... snip
    # the rest here is assembling the event payload
    @glean._record(payload)
  end
end
```

This is very rough and from the top of my head and untested. I'm open for suggestions.

Once we settled on the API we want implementation should be more straight forward.

glean_parser/templates/ruby_server.jinja2
braunreyes: it is not guaranteed, so I am removing this.  the value of the OS version is not important enough to warrant to complex logic.

glean_parser/templates/ruby_server.jinja2
braunreyes: also removing per above convos

glean_parser/templates/ruby_server.jinja2
braunreyes: ok

glean_parser/templates/ruby_server.jinja2
braunreyes: I do not fully understand this question, but I think what you are saying is that we should be sending these logs through whatever logger Mastodon Rails App is using for other types of logs versus hard-coding to stdout.  I can look into this.

glean_parser/templates/ruby_server.jinja2
badboy: oh hehe, I thought you were a ruby guy.

glean_parser/templates/ruby_server.jinja2
akkomar: > (...) we should be sending these logs through whatever logger Mastodon Rails App is using for other types of logs (...)

Yes, best to log to the same destination that application does, but to be fair stdout is probably fine.

braunreyes:
@badboy @akkomar  I pushed my lastest version of the Ruby Server parsing logic.  I checked with Marion(@ml8) on the Social Platform team and she felt your example was reasonable, so I went that route.  I will send this comment her way for further review of the approach.

Here is a sample from the mastodon logs:

image

```json
{"Timestamp":"1697228395","Logger":"glean","Type":"glean-server-event","Severity":"INFO ","Pid":"999","Fields":{"document_namespace":"moso-mastodon","document_type":"events","document_version":"1","document_id":"b7a24050-a9c5-4418-b718-3be525028d5c","user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/118.0","ip_address":"172.19.0.1","payload":"{\"client_info\":{\"telemetry_sdk_build\":\"glean_parser v0.1.dev1004+g4822435\",\"first_run_date\":\"2023-10-13T20:16:56.328Z\",\"os\":\"Linux\",\"os_version\":\"Unknown\",\"architecture\":\"Unknown\",\"app_build\":\"Unknown\",\"app_display_version\":\"4.1.4\",\"app_channel\":\"development\"},\"ping_info\":{\"seq\":16,\"start_time\":\"2023-10-13T20:19:55.555Z\",\"end_time\":\"2023-10-13T20:19:55.555Z\"},\"metrics\":{\"string\":{\"identifiers.adjust_device_id\":null,\"identifiers.fxa_account_id\":null,\"identifiers.mastodon_account_handle\":\"braunzilla@mozilla.social\",\"identifiers.mastodon_account_id\":111229554497968442,\"identifiers.user_agent\":\"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/118.0\"}},\"events\":[{\"category\":\"backend\",\"name\":\"object_update\",\"timestamp\":\"2023-10-13T20:19:55.555Z\",\"extra\":[{\"key\":\"object_type\",\"value\":\"api_request\"},{\"key\":\"object_state\",\"value\":\"{\\\"user_id\\\":2,\\\"path\\\":\\\"/api/v1/lists\\\",\\\"controller\\\":\\\"lists\\\",\\\"method\\\":\\\"GET\\\",\\\"status_code\\\":200}\"}]}]}"}}
```

Here is my place holder PR for showing what this could look like in the Mastodon code base.

https://github.com/MozillaSocial/mastodon/pull/36/files

glean_parser/ruby_server.py
```diff
@@ -0,0 +1,138 @@
...
+) -> None:
+    """
+    Given a tree of objects, output ruby code to `output_dir`.
+
+    :param objects: A tree of objects (metrics and pings) as returned from
+        `parser.parse_objects`.
+    :param output_dir: Path to an output directory to write to.
+    """
+
+    output("ruby", objs, output_dir)
```
badboy, line 138: This is a minimal wrapper around `output` and just passes on the static `"ruby"` string which is unsed.
You can remove this and rename the above function to `output_ruby` instead.

```suggestion
```

tests/test_ruby_server.py
```diff
@@ -0,0 +1,87 @@
...
+        tmpdir,
+    )
+    captured = capsys.readouterr()
+    assert all(False for _ in tmpdir.iterdir())
+    assert "no events found...at least one event metric is required" in captured.out
+
+
+def test_parser_rb_server_metrics_unsupported_type(tmpdir, capsys):
+    """Test that no files are generated if only metric definitions
+    are provided without pings."""
```
badboy, line 50: This doc string seems incorrect, given the test name and test file.

tests/data/ruby_server_metrics_unsupported.yaml
```diff
@@ -0,0 +1,49 @@
...
+  fxa_account_id:
+    type: boolean
+    description: >
+      The user's FxA account ID, if available.
+    bugs:
+      - TBD
+    data_reviews:
+      - TBD
+    notification_emails:
+      - kdemtchouk@mozilla.com
```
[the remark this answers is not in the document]
braunreyes: fixed

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,201 @@
...
+        'ip_address' => ip_address,
+        'payload' => serialized_event_payload,
+      }
+      @logger.formatter = proc do |severity, datetime, _progname, msg|
+        date_format = datetime.to_i
+        logger_name = 'glean'
+        "#{JSON.dump(Timestamp: date_format.to_s, Logger: logger_name.to_s, Type: GLEAN_EVENT_MOZLOG_TYPE.to_s, Severity: severity.ljust(5).to_s, Pid: Process.pid.to_s, Fields: msg)}\n"
+      end
+      @logger.info(ping)
+      return unless @initial_seq == 1_000_000_000
```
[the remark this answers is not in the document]
braunreyes: this means 1 billion...its what rubocop formatted it to

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,201 @@
...
+
+      @initial_seq = 0
+      @first_run_date = t.utc
```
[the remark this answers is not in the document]
braunreyes: I am trying to protect us from hitting the max Ruby int....just being extra safe even though we may never hit it.  Just seemed like the right thing to do.

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,201 @@
...
+        'document_id' => SecureRandom.uuid,
+        'user_agent' => user_agent,
+        'ip_address' => ip_address,
+        'payload' => serialized_event_payload,
+      }
+      @logger.formatter = proc do |severity, datetime, _progname, msg|
+        date_format = datetime.to_i
+        logger_name = 'glean'
+        "#{JSON.dump(Timestamp: date_format.to_s, Logger: logger_name.to_s, Type: GLEAN_EVENT_MOZLOG_TYPE.to_s, Severity: severity.ljust(5).to_s, Pid: Process.pid.to_s, Fields: msg)}\n"
+      end
```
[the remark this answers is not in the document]
braunreyes: moved to `inititialize`

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,201 @@
...
+          'first_run_date' => @first_run_date,
+          'os' => OS.name,
+          'os_version' => 'Unknown',
+          'architecture' => 'Unknown',
+          'app_build' => 'Unknown',
+          'app_display_version' => @app_display_version,
+          'app_channel' => @app_channel,
+        },
+        'ping_info' => {
+          'seq' => @initial_seq += 1,
```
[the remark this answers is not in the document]
braunreyes: done

glean_parser/templates/ruby_server.jinja2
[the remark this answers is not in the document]
braunreyes: @badboy let me know if the above explanation makes sense

glean_parser/templates/ruby_server.jinja2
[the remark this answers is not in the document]
braunreyes: @badboy let me know if the above explanation makes sense

glean_parser/ruby_server.py
```diff
@@ -0,0 +1,138 @@
...
+
+"""
+Outputter to generate server ruby code for collecting events.
+
+This outputter is different from the rest of the outputters in that the code it
+generates does not use the Glean SDK. It is meant to be used to collect events
+using "events as pings" pattern in server-side environments. In these environments
+SDK assumptions to measurement window and connectivity don't hold.
+Generated code takes care of assembling pings with metrics, serializing to messages
+conforming to Glean schema, and logging with mozlog. Then it's the role of the ingestion
```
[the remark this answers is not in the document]
braunreyes: fixed

glean_parser/ruby_server.py
```diff
@@ -0,0 +1,138 @@
...
+Warning: this outputter supports limited set of metrics,
+see `SUPPORTED_METRIC_TYPES` below.
+"""
+from collections import defaultdict
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+from . import __version__, metrics, util
+
+# Adding a metric here will require updating the `generate_js_metric_type` function
```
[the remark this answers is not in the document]
braunreyes: removed the comment

glean_parser/ruby_server.py
```diff
@@ -0,0 +1,138 @@
...
+            if isinstance(metric, metrics.Metric):
+                if metric.type not in SUPPORTED_METRIC_TYPES:
+                    print(
+                        "❌ Ignoring unsupported metric type: "
+                        + f"{metric.type}:{metric.name}."
+                        + " Reach out to Glean team to add support for this"
+                        + " metric type."
+                    )
+                    continue
+                for ping in metric.send_in_pings:
```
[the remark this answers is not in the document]
braunreyes: added logic to check for this

glean_parser/ruby_server.py
```diff
@@ -0,0 +1,138 @@
...
+        print("❌ no events found...at least one event metric is required")
+        return
+    extension = ".rb"
+    filepath = output_dir / ("server_events" + extension)
+    with filepath.open("w", encoding="utf-8") as fd:
+        fd.write(
+            template.render(
+                parser_version=__version__,
+                pings=ping_to_metrics,
+                lang=lang,
```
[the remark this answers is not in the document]
braunreyes: fixed

glean_parser/ruby_server.py
braunreyes: implemented the changes you suggested and updated tests to cover

tests/test_ruby_server.py
braunreyes: fixed all test docstrings

glean_parser/templates/ruby_server.jinja2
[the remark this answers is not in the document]
badboy: The formatting I understand.
My question is: Why are we checking for this overflow? If we _need_ this we should pull it out into a constant so we can give it a name.

Ruby is able to use integers far above 2 billion. Ruby's integers actually automatically extend to a `BigInt` construct able to represent arbitrary numbers.
JSON technically doesn't have integers, but only floats, so at a minimum integers up to 2^53 are reliable, or 2^64 when the other side correctly deserializes it as 64-bit integers.
In any case that are _huge_ numbers and I wouldn't worry about overflow in the Ruby code really.
Sure, we could hit that, but given that we don't store this value on disk anyway it's more likely that the service will be restarted not too far out, thus leading to starting from `seq=0` again anyway. 
All in all sequence numbers for the backend will be less reliable.

glean_parser/templates/ruby_server.jinja2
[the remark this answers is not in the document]
braunreyes: removed this logic per the comment thread above

badboy (changes requested):
Unfortunately this requires one more round.

To test against the schema locally I did this:

Create a file `log.rb` with this content:

```ruby
require_relative 'tmp/server_events'

include Glean

glean = GleanEventsLogger.new(app_id: "test", app_display_version: "0.0.1", app_channel: "nightly", logger_options: STDOUT)

glean.backend_object_update.record(
    object_type: "type",
    object_state: "state",
    identifiers_fxa_account_id: nil,
    user_agent: "glean/test",
    ip_address: "0.0.0.0"
)
```

Then translate the definitions to Ruby:

```
python -m glean_parser translate -f ruby_server -o tmp tests/data/ruby_server_metrics.yaml
```

And finally run the ruby script and validate its output against the schema:

```
ruby log.rb | jq .Fields.payload -r | python -m glean_parser check -s https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/main/schemas/glean/glean/glean.1.schema.json
```

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,197 @@
...
+         {% for metric_type, metrics in metrics_by_type.items() %}
+         {% if metric_type != 'event' %}
+          '{{ metric_type }}' => {
+          {% for metric in metrics %}
+            '{{ metric|metric_name }}' => {{ metric|metric_argument_name }},
+          {% endfor %}
+          },
+        {% endif %}
+        {% endfor %}
+        },
```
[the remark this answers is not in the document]
braunreyes: I have implemented the filtering logic and validated using your functional testing methods

badboy (approved):

glean_parser/templates/ruby_server.jinja2
```diff
@@ -0,0 +1,202 @@
...
+        {% for metric in metrics %}
+          '{{ metric|metric_name }}' => {{ metric|metric_argument_name }},
+        {% endfor %}
+        },
+        {% endif %}
+        {% endfor %}
+      }
+      # filter out key value pairs where value is nil
+      metrics_raw.each do |key, value|
+        metrics_raw[key] = value.compact.transform_values(&:to_s)
```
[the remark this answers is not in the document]
badboy: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1860015 for that

badboy:
rebased & force-pushed. When green I will squash-merge this.

PR #625, opened 2023-09-13
Bug 1851999 - Prevent `ping` lifetime metrics on the `events` ping

perrymcmanis144:

### DIFF
--- a/glean_parser/lint.py
+++ b/glean_parser/lint.py
@@ -286,6 +286,23 @@ def check_old_event_api(
         yield ("The old event API is gone. Extra keys require a type.")
 
 
+def check_metric_on_events_lifetime(
+    metric: metrics.Metric, parser_config: Dict[str, Any]
+) -> LintGenerator:
+    """A non-event metric on the Events ping only makes sense if its value
+    is immutable over the life of the ping."""
+    if (
+        "events" in metric.send_in_pings
+        and "all_pings" not in metric.send_in_pings
+        and metric.type != "event"
+        and metric.lifetime == metrics.Lifetime.ping
+    ):
+        yield (
+            "Non-event metrics sent on the Events ping should not have the ping"
+            " lifetime."
+        )
+
+
 def check_redundant_ping(
     pings: pings.Ping, parser_config: Dict[str, Any]
 ) -> LintGenerator:
@@ -366,6 +383,7 @@ METRIC_CHECKS: Dict[
     "USER_LIFETIME_EXPIRATION": (check_user_lifetime_expiration, CheckType.warning),
     "EXPIRED": (check_expired_metric, CheckType.warning),
     "OLD_EVENT_API": (check_old_event_api, CheckType.warning),
+    "METRIC_ON_EVENTS_LIFETIME": (check_metric_on_events_lifetime, CheckType.error),
 }
 
 
--- a/tests/test_lint.py
+++ b/tests/test_lint.py
@@ -181,8 +181,16 @@ def test_misspelling_pings():
         {
             "user_data": {
                 "counter": {"type": "counter", "send_in_pings": ["metric"]},
-                "string": {"type": "string", "send_in_pings": ["event"]},
-                "string2": {"type": "string", "send_in_pings": ["metrics", "events"]},
+                "string": {
+                    "type": "string",
+                    "lifetime": "application",
+                    "send_in_pings": ["event"],
+                },
+                "string2": {
+                    "type": "string",
+                    "lifetime": "application",
+                    "send_in_pings": ["metrics", "events"],
+                },
             }
         }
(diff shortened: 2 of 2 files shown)


perrymcmanis144:
hey @badboy here's a draft I'm working on. I figured this one looked like it needed to be solved in the python and not a schema so I did it that way.

just as a point of interest, I tried to run `make install-kotlin-linters` and this appears to require an explicit install of coreutils on mac.

badboy:
> just as a point of interest, I tried to run `make install-kotlin-linters` and this appears to require an explicit install of coreutils on mac.

probably because it uses `sha256sum`? We could change it to `shasum -a 256`

badboy:
Can we implement this as a lint? https://github.com/mozilla/glean_parser/blob/b03df4dbced33d84751135ecb34e9696332228d6/glean_parser/lint.py#L357

perrymcmanis144:
> Can we implement this as a lint?
> 
> https://github.com/mozilla/glean_parser/blob/b03df4dbced33d84751135ecb34e9696332228d6/glean_parser/lint.py#L357

Sure, updated to reflect.

glean_parser/lint.py
```diff
@@ -286,6 +286,22 @@ def check_old_event_api(
...
+    """A non-event metric on the events ping only makes sense if its lifetime
+    is tied to that events ping's lifetime."""
+    if (
+        "events" in metric.send_in_pings
+        and "all_pings" not in metric.send_in_pings
+        and metric.type != metrics.Event
+        and metric.lifetime != metrics.Lifetime.ping
+    ):
+        yield (
+            "Non-event metrics sent on the Events ping should have the ping lifetime"
```
badboy, line 301: We usually refer to them by their lowercase name.
```suggestion
            "Non-event metrics sent on the events ping should have the ping lifetime"
```

tests/test_lint.py
```diff
@@ -279,6 +279,27 @@ def test_expires_too_far_in_the_future():
     assert set(["EXPIRATION_DATE_TOO_FAR"]) == set(v.check_name for v in nits)
 
 
+def test_invalid_lifetime_for_metric_on_events_ping():
+    """Test that a `expires` dates too far in the future generates warnings"""
```
badboy, line 283: Copy-pasta. Needs an update.

glean_parser/lint.py
```diff
@@ -286,6 +286,22 @@ def check_old_event_api(
...
+def check_metric_on_events_lifetime(
+    metric: metrics.Metric, parser_config: Dict[str, Any]
+) -> LintGenerator:
+    """A non-event metric on the events ping only makes sense if its lifetime
+    is tied to that events ping's lifetime."""
+    if (
+        "events" in metric.send_in_pings
+        and "all_pings" not in metric.send_in_pings
+        and metric.type != metrics.Event
+        and metric.lifetime != metrics.Lifetime.ping
```
badboy, line 298: Wait, we want to _exclude_ ping-lifetime. `application` and user` lifetime are ok, these are static over time.

glean_parser/lint.py
badboy: As said above, `ping` lifetime should be disallowed, `application` is what should be prefered.

tests/test_lint.py
```diff
@@ -279,6 +279,27 @@ def test_expires_too_far_in_the_future():
...
+    contents = [
+        {
+            "user_data": {
+                "invalid_lifetime": {
+                    "type": "counter",
+                    "lifetime": "application",
```
badboy, line 289: As said this should pass, `ping` should make it fail

badboy (changes requested):

glean_parser/lint.py
```diff
@@ -286,6 +286,23 @@ def check_old_event_api(
...
 
+def check_metric_on_events_lifetime(
+    metric: metrics.Metric, parser_config: Dict[str, Any]
+) -> LintGenerator:
+    """A non-event metric on the Events ping only makes sense if its value
+    is immutable over the life of the ping."""
+    if (
+        "events" in metric.send_in_pings
+        and "all_pings" not in metric.send_in_pings
+        and metric.type != metrics.Event
```
badboy, line 297: `metric.type` is a string, we can't compare that with a class.
```suggestion
        and metric.type != "event"
```

or (I think?)

```suggestion
        and not isinstance(metric, metrics.Event)
```

the former is more stringy, but IMO reads better

badboy (changes requested):
See inline comments for one.
There are more errors after that because in our test data we seem to have a couple of metrics that trigger that linter and thus now fail. The proper way is to make them use either another metric type in the test or set the correct lifetime

badboy (approved):

PR #631, opened 2023-10-10
Bug 1857924 - Add an integration-like test for the javascript_server logger

badboy:
This is what I had in mind for testing the JavaScript server part behaves as expected.
It runs an actual script using `node`, logging to stdout and checking that output against the schema.

### DIFF
--- a/tests/test_javascript_server.py
+++ b/tests/test_javascript_server.py
@@ -4,9 +4,14 @@
 # http://creativecommons.org/publicdomain/zero/1.0/
 
 from pathlib import Path
+import io
+import pytest
+import shutil
+import subprocess
 
 from glean_parser import javascript_server
 from glean_parser import translate
+from glean_parser import validate_ping
 
 
 ROOT = Path(__file__).parent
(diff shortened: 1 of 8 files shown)


tests/test_javascript_server.py
```diff
@@ -67,3 +72,61 @@ def test_generate_ping_factory_method():
...
+            ROOT / "data" / "fxa-server-pings.yaml",
+            ROOT / "data" / "fxa-server-metrics.yaml",
+        ],
+        "javascript_server",
+        tmpdir,
+    )
+
+    factory = "createAccountsEventsEvent"
+    code = """
+events.record({ user_agent: "glean-test/1.0", event_name: "testing" });
```
badboy, line 117: Oh just realized the variable name here is still hardcoded, so we might want to shuffle around what code needs to be provided a bit. Or if we really just stick to one test here we don't bother with templates and just include the full file.

badboy:
> This requires developers to have node installed locally in order for `make test` to pass. Is that OK? Alternative would be to run the test program in a container but that would require more changes to test harness.

IMO yes. We can also hide that behind an additional flag (the schema tests are already behind `--run-web-tests`)

> It might be also worth validating other fields in message envelope ([1], [2]) that ingestion pipeline uses. I can add these here if you want.

That makes sense

tests/test_javascript_server.py
```diff
@@ -67,3 +73,70 @@ def test_generate_ping_factory_method():
...
+    """
+
+    shutil.copy(ROOT / "test-js" / "package.json", code_dir)
+    subprocess.check_call(["npm", "install"], cwd=code_dir)
+
+    tmpl_code = ""
+    with open(ROOT / "test-js" / "test.js.tmpl", "r") as fp:
+        tmpl_code = fp.read()
+
+    tmpl_code = (
```
travis79, line 91: DIY templates? Clever!

travis79 (approved):
This looks fine to me from my perspective

PR #637, opened 2023-11-06
DENG-1736 Add support for event metric type in server-side JavaScript outputter

akkomar:
@badboy this changes js server outputter to support event metric types, motivated by discussion with Moso: https://mozilla-hub.atlassian.net/browse/DENG-1736

Sample code for easier viewing:
* generated class: https://github.com/akkomar/server-telemetry-node-poc/blob/main/test-js-logger/glean-server-event-metric/server_events.js
* usage: https://github.com/akkomar/server-telemetry-node-poc/blob/main/test-js-logger/app.js#L84-L99

This is similar to Ruby server template, however I'd love a critical look on the generated API and its consistency with SDK.


glean_parser/javascript_server.py
```diff
@@ -28,11 +38,14 @@
...
+SUPPORTED_METRIC_TYPES = ["string", "event"]
 
 
-def event_class_name(pingName: str) -> str:
-    return util.Camelize(pingName) + "ServerEvent"
+def event_class_name(pingName: str, event_metric_exists: bool) -> str:
+    # For compatibility with FxA codebase we don't want to add "Logger" suffix
+    # when custom pings without event metrics are used.
+    suffix = "Logger" if event_metric_exists else ""
+    return util.Camelize(pingName) + "ServerEvent" + suffix
```
badboy, line 48: Uff, seeing both camelCase and snake_case mixed here. Should we clean this up while we're here?

```suggestion
def event_class_name(ping_name: str, event_metric_exists: bool) -> str:
    # For compatibility with FxA codebase we don't want to add "Logger" suffix
    # when custom pings without event metrics are used.
    suffix = "Logger" if event_metric_exists else ""
    return util.Camelize(ping_name) + "ServerEvent" + suffix
```

glean_parser/javascript_server.py
```diff
@@ -79,24 +100,16 @@ def output(
...
-    # each metric as in standard outputters.
-    PING_METRIC_ERROR_MSG = (
-        " Server-side environment is simplified and this"
-        + " parser doesn't generate individual metric files. Make sure to pass all"
-        + " your ping and metric definitions in a single invocation of the parser."
-    )
-    if "pings" not in objs:
-        print("❌ No ping definition found." + PING_METRIC_ERROR_MSG)
-        return
+    EVENT_METRIC_EXISTS = False
```
badboy, line 112: Usually all uppercase is for constants. But you're modifying this variable later.
It's only used within this function anyway, so let's turn it to lowercase

```suggestion
    event_metric_exists = False
```

glean_parser/javascript_server.py
```diff
@@ -113,11 +126,39 @@ def output(
                         + " metric type."
                     )
                     continue
+                if metric.type == "event":
+                    # This is used in the template - generated code is slightly
+                    # different when event metric type is used.
+                    EVENT_METRIC_EXISTS = True
```
badboy, line 132:
```suggestion
                    event_metric_exists = True
```

glean_parser/javascript_server.py
```diff
@@ -113,11 +126,39 @@ def output(
...
+    PING_METRIC_ERROR_MSG = (
+        " Server-side environment is simplified and this"
+        + " parser doesn't generate individual metric files. Make sure to pass all"
+        + " your ping and metric definitions in a single invocation of the parser."
+    )
+    if "pings" not in objs:
+        # If events are meant to be sent in custom pings, we need to make sure they
+        # are defined. Otherwise we won't have destination tables defined and
+        # submissions won't pass validation at ingestion.
+        if EVENT_METRIC_EXISTS:
```
badboy, line 147:
```suggestion
        if event_metric_exists:
```

glean_parser/javascript_server.py
```diff
@@ -129,6 +170,7 @@ def output(
             template.render(
                 parser_version=__version__,
                 pings=ping_to_metrics,
+                event_metric_exists=EVENT_METRIC_EXISTS,
```
badboy, line 173:
```suggestion
                event_metric_exists=event_metric_exists,
```

tests/test-js/test_event_metric.js.tmpl
```diff
@@ -0,0 +1,29 @@
+import { createEventsServerEventLogger } from "./server_events.js";
```
badboy, line 1: This file shouldn't be necessary I think?

badboy (changes requested):

tests/test-js/test_event_metric.js.tmpl
akkomar: Right, I eventually used your template with a slight modification, committed this by accident.

tests/test_javascript_server.py
```diff
@@ -140,3 +149,51 @@ def test_run_logging(tmpdir):
...
+
+
+@pytest.mark.node_dependency
+def test_logging_events_ping_with_event_metrics(tmpdir):
+    tmpdir = "/tmp/glean-js-server"
+    tmpdir = Path(str(tmpdir))
+
+    translate.translate(
+        [
+            ROOT / "data" / "ruby_server_metrics.yaml",
```
akkomar, line 161: I think it would also make sense to rename `ruby_server_metrics.yaml` to something like `server_metrics_with_event.yaml`. This definition isn't Ruby-specific and after this PR will be used in js test.

badboy: It would

badboy (approved):

tests/test_javascript_server.py
akkomar: Let me do that before merging.

PR #645, opened 2023-12-19
Go server support

dmueller:
### Problem
We're getting a Go service up and running and want to be able to get some business metrics that will help inform our decisions and awareness of how the ads funnel is performing. Some of these metrics will correlate with client telemetry, so we want to be able to collect them and analyze them in BQ for convenience (to avoid having to compare data from service monitoring dashboard and then have difficulty correlating them, or worry about having different retention durations between data in different systems)

### Solution
Following the patterns of ruby_server and js_server, this adds support for go_server. It only supports the "events" ping and doesn't support custom pings -- this is a difference between ruby_server and js_server, but it was an intentional decision to leave it out from ruby_server and we want that to be the pattern going forward for server implementations.

The go_server implementation supports all three currently allowed types in the extra_keys field for event metrics, and supports string and quantity metrics as well. We expect numeric data to be common (e.g. how many ads were returned to the client), so that is another difference between go_server and the other two existing ones.

The generated code hasn't yet been deployed in GCP to confirm the flow in its entirety, so there may be some follow up adjustments with that step.


glean_parser/templates/go_server.jinja2
```diff
@@ -0,0 +1,263 @@
...
+
+// log type string used to identify logs to process in the Moz Data Pipeline
+var gleanEventMozlogType string = "glean-server-event"
+
+// interface to send the log messages, accepting a call passing in the eventType and the message.
+// TODO: confirm what calls to cloud logging would be required to see if the log method signature makes sense
+// TODO: confirm if this should depend on https://github.com/mozilla-services/go-mozlog
+type Logger interface {
+	log(eventType string, message string)
+}
```
dmueller, line 26: I learned go-mozlog existed on Friday and wondered if that would be the expectation to use -- the JS implementation depends on a mozlog module, wheras the ruby one doesn't; but it does include fields similar to https://github.com/mozilla-services/go-mozlog/blob/master/mozlog.go#L70

It also isn't clear to me yet if would emit the logs to the right place for ingestion in the data pipeline. Does output to Stdout get collected in cloud logs by default?

glean_parser/templates/go_server.jinja2
```diff
@@ -0,0 +1,263 @@
...
+{% if ping != "events" %}
+type {{ ping|ping_type_name }} struct {
+  {% for metric_type, metrics in metrics_by_type.items() %}
+  {% if metric_type != 'event' %}
+  {% for metric in metrics %}
+  {{ metric|metric_argument_name }} {{ metric.type|go_metric_type }} // {{ metric.description|clean_string }}
+  {% endfor %}
+  {% endif %}
+  {% endfor %}
+	{# TODO: if custom pings need to support including any event metric #}
```
dmueller, line 171: I noticed in the JS code that the custom pings logging method includes a param for passing in events, but if we're sending each event individually on their own in the server implementation I wasn't sure if it is really needed to send a custom ping and an event(s) at the same time.

glean_parser/templates/go_server.jinja2
```diff
@@ -0,0 +1,263 @@
...
+		AppDisplayVersion: g.AppDisplayVersion,
+		AppChannel:        g.AppChannel,
+	}
+}
+
+func createPingInfo() pingInfo {
+	// TODO: confirm expected date format in logs,
+	// js uses Date().toISOString(), ex. "2023-12-13T20:21:55.390Z"
+	// ruby uses Time.now.utc, ex. "2023-12-13 20:21:02 UTC"
+	var now = time.Now().UTC().Format("2006-01-02 15:04:05 MST")
```
dmueller, line 109: Is there an expected date format string in the data pipeline, or will either of these work?

glean_parser/templates/go_server.jinja2
akkomar: JS implementation uses mozlog because FxA (first customer) was using it. I'm not sure what's the standard approach for logging for new services now, but I'd say there's no need to bring in mozlog just for telemetry, it's fine if we just structure the log the same way as json.

> It also isn't clear to me yet if would emit the logs to the right place for ingestion in the data pipeline. Does output to Stdout get collected in cloud logs by default?

Yes, they should be if you'll be running in GCPv2 which I believe will be the case.

glean_parser/templates/go_server.jinja2
```diff
@@ -0,0 +1,263 @@
...
+type RequestInfo struct {
+	UserAgent string
+	IpAddress string
+}
+
+// TODO: verify with data team what defaults make sense for user agent / ip address if they are not supplied
+var defaultRequestInfo = RequestInfo{
+	UserAgent: "Unknown",
+	IpAddress: "Unknown",
+}
```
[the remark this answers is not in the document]
akkomar: We should omit these from the logged message.

For context: we are sending some `Unknown` values because these fields are [required by the Glean schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/8004ddad52e2c27b0381afa383cbf0779e465d7c/schemas/glean/glean/glean.1.schema.json#L80-L88).

glean_parser/templates/go_server.jinja2
akkomar: This should be ISO 8601 -formatted string.
The Ruby case you're mentioning is interesting, I see it properly formatted in BQ (e.g. `2023-12-18T06:40:55.143Z`).

glean_parser/templates/go_server.jinja2
dmueller: We chatted on slack and it is preferred that server implementations only support the events ping. The JS implementation supported either and then the ruby implementation only supported events, but that was an intentional decision which we should keep consistent with for go.

It simplifies the implementations and eases querying of the data, since each custom ping could include any of the event metrics defined, but the server implementations by default are sending the event metrics in the events ping.

As we define what data we want to collect, we can discuss if this causes any hurdles.

glean_parser/go_server.py
```diff
@@ -0,0 +1,147 @@
...
+    # In this environment we don't use a concept of measurement window for collecting
+    # metrics. Only "events as pings" are supported.
+    PING_METRIC_ERROR_MSG = (
+        " Server-side environment is simplified and only supports the events ping type."
+        + " You should not be including pings.yaml with your parser call"
+        + " or referencing any other pings in your metric configuration."
+    )
+    if "pings" in objs:
+        print("❌ Ping definition found." + PING_METRIC_ERROR_MSG)
+        return
```
akkomar, line 105: This seems to be copied from Ruby parser, it will work but we might want to revisit this check soon.

A side effect of using a non-declared `events` ping in server apps is that we don't see it in Glean Dictionary (SDK-based apps have it from the core dependency). How to handle that is more of an internal Glean/tooling discussion.

akkomar (approved):
r+wc

LGTM, I have successfully tested this end to end in a staging project.

glean_parser/go_server.py
dmueller: That makes sense, and if we want to declare the events ping in these server environments we could update this check to see what pings are defined and then communicate/fail if there are pings defined other than events.

pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
 markers =
     web_dependency: mark a test that requires a web connection.
+    node_dependency: mark a test that requires node.
+    ruby_dependency: mark a test that requires ruby.
+    go_dependency: mark a test that requires go.
```
badboy, line 6: thank you

tests/test_go_server.py
```diff
@@ -0,0 +1,189 @@
...
+
+    return subprocess.check_output(["go", "run", "test.go"], cwd=code_dir).decode(
+        "utf-8"
+    )
+
+
+@pytest.mark.go_dependency
+def test_run_logging(tmpdir):
+    tmpdir = Path(str(tmpdir))
+    glean_module_path = Path(str(tmpdir) + "/glean")
```
badboy, line 144:
```suggestion
    glean_module_path = tmpdir / "glean"
```

glean_parser/go_server.py
badboy: Can you [file an issue](https://bugzilla.mozilla.org/enter_bug.cgi?assigned_to=nobody%40mozilla.org&bug_ignored=0&bug_severity=normal&bug_status=NEW&bug_type=defect&cf_fx_iteration=---&cf_fx_points=---&cf_status_firefox100=---&cf_status_firefox101=---&cf_status_firefox99=---&cf_status_firefox_esr91=---&cf_tracking_firefox100=---&cf_tracking_firefox101=---&cf_tracking_firefox99=---&cf_tracking_firefox_esr91=---&component=Glean%3A%20SDK&contenttypemethod=list&contenttypeselection=text%2Fplain&defined_groups=1&filed_via=standard_form&flag_type-4=X&flag_type-607=X&flag_type-721=X&flag_type-737=X&flag_type-799=X&flag_type-800=X&flag_type-803=X&flag_type-936=X&flag_type-947=X&form_name=enter_bug&maketemplate=Remember%20values%20as%20bookmarkable%20template&op_sys=Unspecified&priority=P3&product=Data%20Platform%20and%20Tools&rep_platform=Unspecified&status_whiteboard=%5Bglean-sdk%3Am%3F%5D&target_milestone=---&version=unspecified) so we don't forget?

glean_parser/templates/go_server.jinja2
```diff
@@ -0,0 +1,218 @@
...
+	PingInfo   pingInfo     `json:"ping_info"`
+	Metrics    metrics      `json:"metrics"`
+	Events     []gleanEvent `json:"events"`
+}
+
+type gleanEvent struct {
+	Category  string                 `json:"category"`
+	Name      string                 `json:"name"`
+	Timestamp int64                  `json:"timestamp"`
+	Extra     map[string]interface{} `json:"extra"`
```
badboy, line 82: event extras are a string->string map. the types (string, bool, quantity) are client-side only and get stringified before sending.

glean_parser/templates/go_server.jinja2
```diff
@@ -0,0 +1,218 @@
...
+		{% for metric in metrics %}
+			"{{ metric|metric_name }}": params.{{ metric|metric_argument_name }},
+		{% endfor %}
+		},
+		{% endif %}
+    {% endfor %}
+	}
+	var extraKeys = map[string]interface{}{
+    {% for extra, metadata in event.extra_keys.items() %}
+		"{{ extra }}": params.{{ extra|event_extra_name }},
```
badboy, line 196: Here you should stringify the extras

badboy (changes requested):
one thing blocking this, once fixed this should be good to go.

glean_parser/go_server.py
dmueller: filed https://bugzilla.mozilla.org/show_bug.cgi?id=1874887

badboy (approved):

PR #653, opened 2024-01-26
MPP-3642 Add Python log outputter

akkomar:


tests/test-py/test.py
akkomar, line 1: I didn't go with the template here for better readability.

glean_parser/python_server.py
akkomar, line 1: I'd like to look into unifying common logic from these server outputters into a single base class.

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,186 @@
...
+            "document_type": "{{ ping }}",
+            "document_version": "1",
+            "document_id": str(uuid4()),
+            "user_agent": user_agent,
+            "ip_address": ip_address,
+            "payload": event_payload_serialized,
+        }
+
+        ping_envelope = {
+            "Timestamp": timestamp,
```
jwhitlock, line 98: Our existing logs uses integer nanoseconds since the epoch (`time.time_ns`), not the ISO format. Does this get consumed by the Glean ingester, or is it just part of the envelope?

jwhitlock (commented):
Thank @akkomar! I re-generated `server_events.py` for my branch, see 

https://github.com/mozilla/fx-private-relay/commit/ca33268935cb11af25b928e7a4caf8d179950d79

Our log line went from:

```json
{
  "Timestamp": 1706301499675805952,
  "Type": "glean",
  "Logger": "fx-private-relay",
  "Hostname": "jwhitlock-MB-M2.local",
  "EnvVersion": "2.0",
  "Severity": 6,
  "Pid": 98165,
  "Fields": {
    "document_namespace": "relay-backend",
    "document_type": "events",
    "document_version": "1",
    "document_id": "59da013f-d330-42dd-bbbf-045acc13d5df",
    "user_agent": "",
    "ip_address": "",
    "payload": "{\"metrics\": {}, \"events\": [{\"category\": \"email\", \"name\": \"generate_mask\", \"extra\": {\"mozilla_accounts_id\": \"8c7a373d-ee1c-4ac7-bafa-cef3c6664f16\", \"is_random_mask\": true, \"created_by_api\": true, \"has_generated_for\": false}, \"timestamp\": 1706301499675}], \"ping_info\": {\"seq\": 0, \"start_time\": \"2024-01-26T20:38:19.675773+00:00\", \"end_time\": \"2024-01-26T20:38:19.675773+00:00\"}, \"client_info\": {\"telemetry_sdk_build\": \"glean_parser v11.0.1\", \"first_run_date\": \"Unknown\", \"os\": \"Unknown\", \"os_version\": \"Unknown\", \"architecture\": \"Unknown\", \"app_build\": \"Unknown\", \"app_display_version\": \"2024.01.17\", \"app_channel\": \"local\"}}",
    "msg": "glean-server-event"
  }
}
```
to
```json
{
  "Timestamp": "2024-01-26T21:00:51.038607+00:00",
  "Logger": "glean",
  "Type": "glean-server-event",
  "Fields": {
    "document_namespace": "relay-backend",
    "document_type": "events",
    "document_version": "1",
    "document_id": "68517f06-4baa-44a1-8129-d727a72a830f",
    "user_agent": "",
    "ip_address": "",
    "payload": "{\"metrics\": {}, \"events\": [{\"category\": \"email\", \"name\": \"generate_mask\", \"extra\": {\"mozilla_accounts_id\": \"8d6bc83a-959f-44d0-b6b6-5121b7fa112f\", \"is_random_mask\": \"True\", \"created_by_api\": \"True\", \"has_generated_for\": \"False\"}, \"timestamp\": 1706302851038}], \"ping_info\": {\"seq\": 0, \"start_time\": \"2024-01-26T21:00:51.038607+00:00\", \"end_time\": \"2024-01-26T21:00:51.038607+00:00\"}, \"client_info\": {\"telemetry_sdk_build\": \"glean_parser v11.0.2.dev5+g5f7d68ea\", \"first_run_date\": \"Unknown\", \"os\": \"Unknown\", \"os_version\": \"Unknown\", \"architecture\": \"Unknown\", \"app_build\": \"Unknown\", \"app_display_version\": \"2024.01.17\", \"app_channel\": \"local\"}}"
  }
}
```

There's some differences in the structure and the details, which may be OK. See comments for details.

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,186 @@
...
+		{% endfor %}
+		{% endif %}
+        """
+        event = {
+            "category": "{{ event.category }}",
+            "name": "{{ event.name }}",
+			{% if event.extra_keys %}
+			"extra": {
+				{% for extra, metadata in event.extra_keys.items() %}
+				"{{ extra }}": str({{ extra }}),
```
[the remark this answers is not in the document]
badboy: In Glean event extra values are always encoded as a string. This is a short-coming of our current schema. Bools are usually lower-case though (`"true"` / `"false"`), as that's how glean-core encodes it.

glean_parser/python_server.py
badboy: yeah, let's do that in a followup.

badboy (approved):
mostly similar implementation to the other server outputters.
The details of the generated Python code I leave up to you for now.
All else good to me!

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,186 @@
...
+
+        ping_envelope = {
+            "Timestamp": timestamp,
+            "Logger": "glean",
+            "Type": GLEAN_EVENT_MOZLOG_TYPE,
+            "Fields": ping
+        }
+        ping_envelope_serialized = json.dumps(ping_envelope)
+
+        print(ping_envelope_serialized)
```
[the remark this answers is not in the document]
akkomar: Extracting this to a method sounds fine to me.
FWIW fields in the ping envelope are inspired by Mozlog which you probably don't use. We don't really need `Timestamp` and `Logger`, but other than that our infrastructure expects the log line to be a JSON with specified `Type` and data in `Fields`.

glean_parser/templates/python_server.jinja2
akkomar: It's not consumed, see also https://github.com/mozilla/glean_parser/pull/653#discussion_r1471524494

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,186 @@
...
+        :param str application_id: The application ID.
+        :param str app_display_version: The application display version.
+        :param str channel: The channel.
+        """
+        self._application_id = application_id
+        self._app_display_version = app_display_version
+        self._channel = channel
+
+    def _record(
+		self,
```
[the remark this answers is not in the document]
akkomar: Done.

glean_parser/templates/python_server.jinja2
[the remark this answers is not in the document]
akkomar: Maybe we should just use `logging` in the template then? That would be similar to what we do in JS/Ts.

It would be something like:
```
import logging
logger = logging.getLogger(GLEAN_EVENT_MOZLOG_TYPE)
log.info(ping_envelope_serialized)
```
We'd require applications to configure the output like you do.
I guess you'd only need to configure this logger in https://github.com/mozilla/fx-private-relay/blob/5659e0ef42be097a152cba1b9153fbac2e3421d1/privaterelay/settings.py#L580-L618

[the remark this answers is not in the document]
jwhitlock: I think `print` is a decent default, requiring no configuration and you've got good test. Relay can override `emit_record` to use the log. We could address the flexibility with documentation.

Alternatively, we could add an initialization option to use logging, or try to read the logging configuration to see if there is an appropriate logger (I'd need to explore if this is possible).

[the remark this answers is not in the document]
jwhitlock: If we're just looking for the dockerflow mozlogger, which may be reasonable for Mozilla Python projects, this works:

```
def is_dockerflow_mozlog() -> bool:

    try:
        from dockerflow.logging import JsonLogFormatter
    except ImportError:
        return False
    import logging
    logger = logging.getLogger(GLEAN_EVENT_MOZLOG_TYPE)
    for handler in logger.handlers:
        if isinstance(handler.formatter, JsonLogFormatter):
            return True
    return False
```

I've pushed https://github.com/mozilla/fx-private-relay/commit/51f5e88643822b7a83e4982df0aea60bc728fb7c to show what the generated code looks like in our repo.

akkomar:
I think this is ready to be merged now. Re-requesting review from @jwhitlock for a final confirmation.

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,194 @@
...
+    def emit_record(self, now: datetime, ping:dict[str, Any]) -> None:
+        """Log the ping to STDOUT.
+        Applications might want to override this method to use their own logging.
+        If doing so, make sure to log the ping as JSON, and to include the
+        `Type: GLEAN_EVENT_MOZLOG_TYPE`."""    
+        ping_envelope = {
+            "Timestamp": now.isoformat(),
+            "Logger": "glean",
+            "Type": GLEAN_EVENT_MOZLOG_TYPE,
+            "Fields": ping
```
jwhitlock, line 109: Nit: `black` adds a comma
```suggestion
            "Fields": ping,
```

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,194 @@
...
+
+        :param str application_id: The application ID.
+        :param str app_display_version: The application display version.
+        :param str channel: The channel.
+        """
+        self._application_id = application_id
+        self._app_display_version = app_display_version
+        self._channel = channel
+
+    def _record(
```
jwhitlock, line 35: Nit: `black` formats this as one line, because there are no metrics. We could special case no metrics, but that may be ugly.

glean_parser/python_server.py
```diff
@@ -0,0 +1,130 @@
...
+            return
+
+    extension = ".py"
+    filepath = output_dir / ("server_events" + extension)
+    with filepath.open("w", encoding="utf-8") as fd:
+        fd.write(template.render(parser_version=__version__, pings=ping_to_metrics))
+
+    # create an empty `__init__.py` file to make the directory a package
+    init_file = output_dir / "__init__.py"
+    with init_file.open("w", encoding="utf-8") as fd:
```
jwhitlock, line 129: Suggestion: This could create it if it doesn't exist. I was using our `glean/__init__.py`, was a little surprised to see it overwritten.

jwhitlock (approved):
Thanks @akkomar, this works for Relay! I've updated [emit-pings-as-logs-mpp-3636](https://github.com/mozilla/fx-private-relay/compare/emit-pings-as-logs-mpp-3636) with a class that overrides `emit_record`, and it emits a JSON log that I believe is compatible with the `print` version:

```json
{
  "Timestamp": 1706718565532915968,
  "Type": "glean-server-event",
  "Logger": "fx-private-relay",
  "Hostname": "jwhitlock-MB-M2.local",
  "EnvVersion": "2.0",
  "Severity": 6,
  "Pid": 21381,
  "Fields": {
    "document_namespace": "relay-backend",
    "document_type": "events",
    "document_version": "1",
    "document_id": "841f8932-653c-4454-8092-6f3f6df254cf",
    "user_agent": "",
    "ip_address": "",
    "payload": "{\"metrics\": {}, \"events\": [{\"category\": \"email\", \"name\": \"generate_mask\", \"extra\": {\"mozilla_accounts_id\": \"e699353f-1cfd-40fe-95e6-9402c42ba5c2\", \"is_random_mask\": \"False\", \"created_by_api\": \"False\", \"has_generated_for\": \"False\"}, \"timestamp\": 1706718565532}], \"ping_info\": {\"seq\": 0, \"start_time\": \"2024-01-31T16:29:25.532888+00:00\", \"end_time\": \"2024-01-31T16:29:25.532888+00:00\"}, \"client_info\": {\"telemetry_sdk_build\": \"glean_parser v11.0.2.dev6+g31512237\", \"first_run_date\": \"Unknown\", \"os\": \"Unknown\", \"os_version\": \"Unknown\", \"architecture\": \"Unknown\", \"app_build\": \"Unknown\", \"app_display_version\": \"unknown\", \"app_channel\": \"local\"}}",
    "msg": "glean-server-event"
  }
}
```

I've added a question about boolean formatting, and about overwriting the existing `__init__.py`, but neither are blockers.

I also have a bunch of nits about how `black` re-formats the file. Take or leave those suggestions, it should be easy enough to tell `black` to ignore generated files, or reformat after generation. If you like the suggestions but don't have time, I'm happy to submit a follow-on PR.

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,194 @@
...
+        {% endfor %}
+        {% endif %}
+        {% endfor %}
+        event: dict[str, Any]
+    ) -> None:
+        now = datetime.now(timezone.utc)
+        timestamp = now.isoformat()
+        event["timestamp"] = int(1000.0 * now.timestamp())  # Milliseconds since epoch
+        event_payload = {
+            "metrics": {
```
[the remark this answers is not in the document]
akkomar: I'd leave this as is for the sake of readability of the template. We can revisit if this turns out inconvenient for consumers.

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,194 @@
...
+            "name": "{{ event.name }}",
+            {% if event.extra_keys %}
+            "extra": {
+                {% for extra, metadata in event.extra_keys.items() %}
+                "{{ extra }}": str({{ extra }}),
+                {% endfor %}
+            },
+            {% endif %}
+        }
+        self._record(
```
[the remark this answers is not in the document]
akkomar: I'd leave this as is for the sake of readability of the template. We can revisit if this turns out inconvenient for consumers.

glean_parser/templates/python_server.jinja2
akkomar: I'd leave this as is for the sake of readability of the template. We can revisit if this turns out inconvenient for consumers.

glean_parser/python_server.py
akkomar: Is https://github.com/mozilla/fx-private-relay/commit/51f5e88643822b7a83e4982df0aea60bc728fb7c#diff-5d5a21ba24b718227a3e2e46a94a9333471dfd62ef7b673fb7b5d7de9e84da8eL1-L3 all that you'd need in the package init? If so, we can generate this too. I'd be inclined to keep the overwrite behavior to keep this consistent with the rest of the glean_parser.

jwhitlock: Yep, but there isn't much different between:

```
from privaterelay.glean import EventsServerEventLogger, create_events_server_event_logger
```

and
```
from privaterelay.glean.server_events import EventsServerEventLogger, create_events_server_event_logger
```

I don't think it is worth the complexity of generating a non-blank `__init__,py` to break consistency with the rest of `glean_parser`. Withdrawing my suggestion.

glean_parser/templates/python_server.jinja2
```diff
@@ -0,0 +1,193 @@
...
+            "document_version": "1",
+            "document_id": str(uuid4()),
+            "user_agent": user_agent,
+            "ip_address": ip_address,
+            "payload": event_payload_serialized,
+        }
+
+
+        self.emit_record(now, ping)
+    def emit_record(self, now: datetime, ping:dict[str, Any]) -> None:
```
jwhitlock, line 99: Oops, my last suggestion lost the line between methods:

```suggestion
        self.emit_record(now, ping)

    def emit_record(self, now: datetime, ping:dict[str, Any]) -> None:
```

jwhitlock (approved):
Thanks for the updates! I've updated my branch with the new generate file, and expect lowercase strings for booleans:

https://github.com/mozilla/fx-private-relay/compare/emit-pings-as-logs-mpp-3636

akkomar:
Yes, this should be merged.

PR #670, opened 2024-03-05
Switch to ruff as linter/formatter

badboy:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,11 +1,5 @@
 [bdist_wheel]
 python_tag = py3
 
-[flake8]
-exclude = docs
-max-line-length = 88
-# W503 - line break before binary operator - black formats it that way
-ignore = W503
-
 [aliases]
 test = pytest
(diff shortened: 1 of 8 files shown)


setup.cfg
```diff
@@ -1,11 +1,5 @@
 [bdist_wheel]
 python_tag = py3
 
-[flake8]
-exclude = docs
-max-line-length = 88
-# W503 - line break before binary operator - black formats it that way
-ignore = W503
```
Dexterp37, line 8: Are these in a toml that was not uploaded?

badboy: max-line length is default, ignore unneeded and the exclude unnecessary, because we explicitly pass the directories in our makefile

Dexterp37 (approved):
@badboy are we going to be ok with vendoring ruff in m-c?

badboy:
> @badboy are we going to be ok with vendoring ruff in m-c?

First: [ruff is in use in m-c](https://searchfox.org/mozilla-central/rev/6b0b8c1003504e302e243c300355ef4d61e24909/tools/lint/python/ruff_requirements.in).
Second: we're not vendoring it. It's just a dev tool, specified in `requirements_dev.txt`, not as a direct dependency of the package.

PR #681, opened 2024-04-02
Bug 1889114 - Add 'enabled' property to pings

travis79:
This is enabled initially for Swift, Kotlin, and Rust

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- BREAKING CHANGE: Expose the optional `enabled` property on pings, defaulting to `enabled: true` ([#681](https://github.com/mozilla/glean_parser/pull/681))
+
 ## 13.0.1
 
 - Use faster C yaml parser if available ([#677](https://github.com/mozilla/glean_parser/pull/677))
--- a/glean_parser/schemas/pings.2-0-0.schema.yaml
+++ b/glean_parser/schemas/pings.2-0-0.schema.yaml
@@ -175,6 +175,18 @@ additionalProperties:
       additionalProperties:
         type: string
 
+    enabled:
+      title: Whether or not this ping is enabled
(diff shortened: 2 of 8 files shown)


glean_parser/schemas/pings.2-0-0.schema.yaml
```diff
@@ -175,6 +175,14 @@ additionalProperties:
       additionalProperties:
         type: string
 
+    enabled:
+      title: Whether or not this ping is enabled
+      description: |
+        When `true`, the ping will be sent as usual.
+        When `false`, the ping will not be sent, but the data will continue to
+        be collected but will not be cleared when the ping is submitted.
```
chutten, line 183: Optional parameters need to have a description that:
1) Says they're optional
2) Says what the default value will be if it's omitted

chutten (approved):
r+ with comments addressed... and with a test. Isn't there supposed to be a test? At least we probably need a ping in `tests/data/pings.yaml` that supplies a value for this.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Expose the optional `enabled` property on pings, defaulting to `enabled: true` ([#681](https://github.com/mozilla/glean_parser/pull/681))
```
badboy, line 5: Note that because of the additional parameter generated in code this is a breaking change.

glean_parser/schemas/pings.2-0-0.schema.yaml
travis79: Yeah, I was just following the existing examples in this file where only the _required_ properties are labeled as such in the description. I like having more detail though so you won't get any arguments from me over adding it here.

travis79:
Test added, including pings with explicit setting of `enabled` to the pings.yaml file

PR #693, opened 2024-04-23
AE-321 add support for datetime go_server.py

gleonard-m:

- [AE-321](https://mozilla-hub.atlassian.net/browse/AE-321) Added support for `datetime` in `go_server.py`


travis79 (approved):

glean_parser/go_server.py
```diff
@@ -58,6 +58,8 @@ def generate_metric_type(metric_type: str) -> str:
         return "string"
     elif metric_type == "boolean":
         return "bool"
+    elif metric_type == "datetime":
+        return "string"
```
badboy, line 62: Why is this a string? How is this going to be used?

tests/test_go_server.py
```diff
@@ -104,75 +98,3 @@ def test_parser_go_server(tmp_path):
...
-    subprocess.call(["go", "mod", "init", "glean"], cwd=code_dir)
-    subprocess.call(["go", "mod", "tidy"], cwd=code_dir)
-
-    return subprocess.check_output(["go", "run", "test.go"], cwd=code_dir).decode(
-        "utf-8"
-    )
-
-
-@pytest.mark.go_dependency
-def test_run_logging(tmp_path):
```
badboy, line 132: Why is this test removed?

badboy (changes requested):

tests/test_go_server.py
gleonard-m: Removed test `test_run_logging`. Test fails due to `sudo apt install golang-go` installing go 1.18 and `slog` is included in 1.22.

glean_parser/go_server.py
gleonard-m: The generated tables will include a column of type `STRING`.  With the `metrics.yaml` file indicating the type of `datetime` the generated view will prefix the existing `STRING` column name with `raw_` and add a new column of type `TIMESTAMP`using the UDF `mozfun.glean.parse_datetime()` to convert the STRING value to the TIEMSTAMP.

As an example of the expected view see the `metrics.datetime` metrics `moz-fx-data-shared-prod.firefox_desktop.spoc`.

dmueller: would it makes sense for the go type to be `Time` or something, so that the user of the glean_parser doesn't need to know what format string glean expects for the datetime fields?

as a go glean_parser user, if I didn't already know what format string should be, how would I figure that out?

gleonard-m: Yep - will make the change.

tests/test_go_server.py
akkomar: We shouldn't remove this as it's the only integration test for the Go parser.
Do we have a standard way to log in Go services? Slog looks like a better way to log here than fmt so I think it's perfectly fine to bump the Go version.
I'm happy to help here, just let me know.

glean_parser/templates/go_server.jinja2
```diff
@@ -157,7 +157,7 @@ func (g GleanEventsLogger) record(
 	if envelopeErr != nil {
 		panic("Unable to marshal log envelope to json")
 	}
-	fmt.Println(string(envelopeJson))
+	slog.Debug(string(envelopeJson))
```
akkomar, line 160: Why Debug? I'd rather leave this at info so it works out of the box for new services. If you prefer to have this logged at debug level we can add a template parameter to change the level case by case.

glean_parser/go_server.py
akkomar: +1, this should be a native timestamp type converted to ISO-8601 string before putting in a ping payload. Compare https://github.com/mozilla/glean_parser/blob/bb3beb7258fce567f953d7ec829d31dca488059e/glean_parser/templates/go_server.jinja2#L108

akkomar (changes requested):
We should keep the integration test.

tests/test_go_server.py
gleonard-m: I'm currently re-working this and reverting to the original `fmt.Println` implementation.  Using slog requires configuration in the main application to output the logs in the correct format needed here.

glean_parser/templates/go_server.jinja2
gleonard-m: Reverted to `fmt.Println`

badboy (approved):

PR #740, opened 2024-08-01
Bug 1906941 - Update kotlin jinja template import statement

travis79:
Now that Glean v60.4.0 has released and `service-glean` android-component has been removed, we can update the Kotlin template to use the glean import of `HistogramBase` which will allow us to do away with the `HistogramMetricBase` alias completely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Fix JS and Ruby server templates to correctly send event extra values as strings ([DENG-4405](https://mozilla-hub.atlassian.net/browse/DENG-4405))
 - ENHANCEMENT: Extra keys in `extra_keys:` fields may now contain any printable ASCII characters ([bug 1910976](https://bugzilla.mozilla.org/show_bug.cgi?id=1910976))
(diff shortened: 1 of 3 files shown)


chutten (approved):

badboy:
Note that this is a breaking change and has to go in a major release, to avoid breaking existing code (which auto-updates to minor releases).
Thus this should have a changelog entry.

travis79:
> Note that this is a breaking change and has to go in a major release, to avoid breaking existing code (which auto-updates to minor releases). Thus this should have a changelog entry.

Is it breaking though if the fix to keep it from breaking is already in place? The breaking part will be removing the alias from glean_core/android ;)

I'll add a changelog entry, regardless. I should have done that anyway :)

badboy:
> > Note that this is a breaking change and has to go in a major release, to avoid breaking existing code (which auto-updates to minor releases). Thus this should have a changelog entry.
> 
> Is it breaking though if the fix to keep it from breaking is already in place? The breaking part will be removing the alias from glean_core/android ;)

Hmmm, I guess right now the same thing is available under both names? If that is so then it might be non-breaking indeed.

PR #743, opened 2024-08-13
Bug 1913129 - Remove Geckoview Streaming helper code

badboy:
This is not in use anymore.
Last use removed in https://bugzilla.mozilla.org/show_bug.cgi?id=1877836

We don't remove support for _parsing_ `gecko_datapoint` to stay compatible with already existing old definition files.


badboy:
~~Actually I think this is not even breaking.~~
~~The only user ever was us in FOG. ~~
~~We removed all of that use in Firefox 129.~~

It is breaking, because ESR 128 is also using glean_parser 14.
So what we probably need to do to unbreak 128/release is:

Revert the removal of the alias. Release that as a patch release.
Then re-revert that, do a major release (and just include this too).

travis79 (approved):

PR #758, opened 2024-10-11
extend go_server outputter to support custom pings

dmueller:
### Problem
https://mozilla-hub.atlassian.net/browse/AE-576

For our server integration in go we would like to use custom pings, as this will give more flexibility into how the data gets organized in BQ when we have more than one ping.

### Solution
Adjust error message scenarios in go_server so that it is OK for pings to be defined, and OK if the events ping is not used. The go_server outputter now suppports only events ping, only custom ping(s), or both.

Added more test yaml configuration for each of these three scenarios, since they emit different methods in the templated code.

Enabled boolean support, since the code was already present to support booleans-as-strings in event extra; so now we will also support boolean metric values as well.

Added "time" to test.go.tmpl so that in the unit tests we can populate the datetime field with a value instead of leaving it with a default.

### Testing
`make test-full`

I haven't run this in a place that would actually ingest the logged output, but have run the code similar to how the unit test runs test.go to see a log statement printed to stdout. This would be helpful to confirm that the statement is formatted as expected.

- [X] **Quality**: This PR builds and tests run cleanly
  - `make test-full` runs without any test failures
- [X] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [X] **Changelog**: This PR includes a changelog entry to `CHANGELOG.md` or an explanation of why it does not need one


akkomar (commented):
Regarding the template: instead of treating `events` ping as special case, can we generalize the current approach to custom pings?
This would reduce code duplication and make Go template consistent with others (https://github.com/mozilla/glean_parser/blob/main/glean_parser/templates/javascript_server.jinja2, https://github.com/mozilla/glean_parser/blob/main/glean_parser/templates/python_server.jinja2)

glean_parser/go_server.py
```diff
@@ -120,29 +127,31 @@ def output_go(
...
+        print("❌ No pings with metrics found." + PING_METRIC_ERROR_MSG)
         return
 
     extension = ".go"
     filepath = output_dir / ("server_events" + extension)
     with filepath.open("w", encoding="utf-8") as fd:
         fd.write(
             template.render(
-                parser_version=__version__, events_ping=ping_to_metrics["events"]
+                parser_version=__version__,
```
efixler, line 153: I noticed that the `server_go*.go` files in tests/data don't have current version, maybe __version__ isn't being set? (They also don't have the leading `v`, so that might be incorrect)

akkomar (approved):
LGTM, my only suggestions are about docs and comments to make them (hopefully) clearer.

> I haven't run this in a place that would actually ingest the logged output, but have run the code similar to how the unit test runs test.go to see a log statement printed to stdout. This would be helpful to confirm that the statement is formatted as expected.

Last two tests in `test_go_server.py` validate output against Glean schema. I have also tested this manually with ingestion Decoder and haven't seen any issues.

glean_parser/go_server.py
dmueller: That's the pattern in the test data output so it's not hardcoded; generating those files to compare to it has something like `glean_parser v15.0.2.dev7+gd4b1b802.d20241016` which is manually replaced with `{current_version}` to create these files.

The tests swap that string with whatever the current version is
```
    glean_version = f"glean_parser v{glean_parser.__version__}"
    # use replace instead of format since Go uses { }
    compare = compare_raw.replace("{current_version}", glean_version)
```

glean_parser/go_server.py
```diff
@@ -32,7 +35,15 @@
 
 # Adding a metric here will require updating the `generate_metric_type` function
 # and require adjustments to `metrics` variables the the template.
-SUPPORTED_METRIC_TYPES = ["string", "quantity", "event", "datetime"]
+SUPPORTED_METRIC_TYPES = ["string", "quantity", "event", "datetime", "boolean"]
+
+
+def generate_ping_type_name(ping_name: str) -> str:
+    return f"Ping{util.Camelize(ping_name)}"
```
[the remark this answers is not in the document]
dmueller: I did consider that, but since the "Event" in generate_event_type_name was a prefix I followed that. Agree that it would read better as a suffix so let me try that out for all three of these.

glean_parser/go_server.py
dmueller: Looking more closely, this originally came from the ruby implementation that I based the go one off of; but none of the other ones have these tests to validate the produced code against a "golden file".. we could consider removing this kind of test, since there are tests which generate the code & perform calls to verify the log output produced.

https://github.com/mozilla/glean_parser/blob/main/tests/test_ruby_server.py#L79-L99

akkomar: IIRC we didn't have integration tests that called the generated code and validated schema back when the Ruby test was added. I think it would be fine to get rid of `_compare` tests but it's fine to keep them too. I was planning to refactor and standardize server parsers at some point so will probably address that then.

PR #772, opened 2024-11-12
feat: add rust server and template code SYNC-4447

taddes:
This feature adds `rust_server` as one of the glean_parser options. As part of our discussions to integrate Glean into syncstorage-rs, this is to accomplish that goal and allow for other projects to make use of the parser. The logic was modeled after the `go_server` with a few minor modifications. Template file generates the metrics and events as expected. Tests were also modeled after the go style. 

There are a set of 3 tests that generate rust code via cargo and they pass locally, while would require a little more CI config and I wanted to discuss that with you prior, since there may be added overhead with rustup and building via cargo. Tests temporarily disabled.


taddes:
Appears when running the added rust tests (just the 3 that temporarily generate Rust projects to compare log output against), the CI fails due to some likely incompatibility with cargo. Think we would need to add it to the docker builds, I can go ahead and give that a shot to see if it resolves.

CHANGELOG.md
```diff
@@ -2,12 +2,13 @@
 
 ## Unreleased
 
+## 15.3.0
```
badboy, line 5: Please don't tag the version just yet.

```suggestion
```

CHANGELOG.md
```diff
@@ -2,12 +2,13 @@
...
+- Add rust_server feature to glean_parser to allow Rust server-side metrics ([#772](https://github.com/mozilla/glean_parser/pull/772))
+
 ## 15.2.1
 
 - Allow earlier versions of platformdirs ([#769](https://github.com/mozilla/glean_parser/pull/769))
 
-## 15.2.0
-
```
badboy, line 10: huh, this should definitely not go away.

Makefile
```diff
@@ -77,3 +77,8 @@ install-kotlin-linters: ## install ktlint and detekt for linting Kotlin output
 	chmod a+x ktlint
 	test -f detekt-cli.jar || curl -sSL --output "detekt-cli.jar" https://github.com/detekt/detekt/releases/download/v1.23.6/detekt-cli-1.23.6-all.jar
 	echo "898dcf810e891f449e4e3f9f4a4e2dc75aecf8e1089df41a42a69adb2cbbcffa *detekt-cli.jar" | shasum -a256 -c -
+
+install-rust: ## install rustup and rust toolchain
+	curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+	echo 'source "$$HOME/.cargo/env"' >> ~/.bashrc  # Add Rust to bash startup
+	. ~/.bashrc  # Reload bashrc to include Rust in PATH
```
badboy, line 84: I don't think we should have something that changes the user's environment as a make task. Let's do that in the CI config only. I can do that.

glean_parser/templates/rust_server.jinja2
```diff
@@ -0,0 +1,308 @@
...
+
+        let envelope: LogEnvelope = LogEnvelope {
+            timestamp: Utc::now().timestamp().to_string(),
+            logger: "glean".to_string(),
+            log_type: GLEAN_EVENT_MOZLOG_TYPE.to_string(),
+            fields: ping,
+        };
+        let envelope_json =
+            serde_json::to_string(&envelope).expect("unable to marshal payload to json.");
+        println!("{}", envelope_json);
```
badboy, line 178: We probably shouldn't just print to stdout?
Not sure what the exact plan was here, but wouldn't we want to integrate with `log`? Or somehow pass a logger in?

badboy (changes requested):
Didn't yet do a full review, but noticed a couple of things early, so submitting them already.

glean_parser/templates/rust_server.jinja2
taddes: I think in the discussions with @akkomar and @travis79 , we had decided stdout is the way to go for rust server metrics, however I could be mistaken. I will double check with them, but I think this is how it's expected to be picked up.

CHANGELOG.md
taddes: Good catch, I think when I rebased it removed that line for some reason and I didn't see. Thx

Makefile
taddes: Sounds good, let me know your preferred method for that, as I've implemented it in a lot of varied ways. Ideally, the CircleCI pre-built images are a good way to go, but it's a python application so that doesn't make sense in this case I imagine.

glean_parser/templates/rust_server.jinja2
akkomar: How do you normally output logs in Sync @taddes? It would be best to use your standard logging mechanism. We just need to make sure that this log goes to Cloud Logging in a correct format.

akkomar: I think it's good to go with the approach recommended by SRE/SecOps in the template. I don't remember why Go template looks this way, maybe that's what they do in their code.

Makefile
badboy: https://github.com/mozilla/glean_parser/pull/773 landed, so you should be able to re-base on top of that and re-enable your tests.

taddes: Awesome @badboy , thank you for landing that! Will do.

akkomar (changes requested):
Thank you for working on this!

I added some suggestions inline. I have tested this with my suggested changes applied in a sandbox project and confirmed that the format is correct.

I still have to review the tests. We could go without adding new test yaml files, but as I mentioned earlier I'm planning to refactor and unify these outputters so this is probably fine.

tests/data/server_events_and_custom_ping_compare.rs
```diff
@@ -0,0 +1,339 @@
...
+    events: Vec<GleanEvent>,
+}
+
+#[derive(Serialize, Deserialize)]
+struct LogEnvelope {
+    timestamp: String,
+    logger: String,
+    #[serde(rename = "type")]
+    log_type: String,
+    fields: Ping,
```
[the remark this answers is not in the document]
taddes: Ah yes, that's right, good point as I do recall the MozLog format requires the uppercase!

glean_parser/templates/rust_server.rs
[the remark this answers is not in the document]
taddes: Ah yes, it was my test file for output and verifying that clippy and the cargo checks were appeased, will do!

tests/data/server_events_and_custom_ping_compare.rs
```diff
@@ -0,0 +1,339 @@
...
+    client_info: ClientInfo,
+    ping_info: PingInfo,
+    metrics: Metrics,
+    events: Vec<GleanEvent>,
+}
+
+#[derive(Serialize, Deserialize)]
+struct LogEnvelope {
+    timestamp: String,
+    logger: String,
```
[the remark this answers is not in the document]
taddes: Sounds good. So to be clear, these fields should not be defined at all within the struct and then when instantiating a `LogEvelope` instance, there will be no `timestamp` or `logger` at all? Thx for the context.

taddes:
Excellent @akkomar , thank you for taking the time to review and also to validate in the sandbox. I will be doing the same with my custom project that emulates the data being passed in through Sync. Given each language varies in its serialization logic and whatnot, please let me know if you find anything that warrants a change. I will be sure to do the same.

I've updated the template to be compliant with MozLog (Type and Fields), removing the two `timestamp` and `logger` fields from the `LogEvelope` struct and updated the test files/tests.

We'll confer next week with any other modifications. Now that we're registered with the probe scraper, once Sync is deployed with the Glean logic we'll be able to verify the functionality.

Thanks for the opportunity to contribute to your project and for all the great collaboration!

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+## 15.3.0
+- Add `rust_server` feature to glean_parser to support Rust server-side metrics ([#772](https://github.com/mozilla/glean_parser/pull/772))
```
akkomar, line 6:
```suggestion
- Add Rust log outputter (`rust_server`) ([#772](https://github.com/mozilla/glean_parser/pull/772))
```

akkomar (commented):
All right, looks like the only unresolved comment we have is about the actual logger:
> We use slog / [slog_mozlog_json](https://github.com/mozilla/glean_parser/pull/github.com/mozilla-services/slog-mozlog-json) per [SecOps guidelines](https://github.com/mozilla-services/websec-check/blob/main/rust.md). I'm going to check with the team and SRE about this.

@taddes have you checked with the team/SRE?
I'd rather have this template to use the default recommended logger that you (and others) use, so you can just drop in the generated code into your repo without modifications.

taddes:
Ok @akkomar , in discussion with the team, we agreed stdout is ideal and we are not going to use `slog`. It would introduce an unnecessary layer that diverges from the desired output. A couple minor changes introduced as specified in my last comment (a few `to_owned()` calls instead of `to_string()` and a `Default` impl for `Pinginfo`

akkomar (approved):
LGTM. I'd love @badboy to take a quick look at the generated code (for convenience [see here](https://github.com/akkomar/server-telemetry-node-poc/pull/1/files#diff-c694e95fd2dc337599b52405253edd0ffe00fa93e870cb34f36358fe17628c88)) before merging.

taddes:
Awesome, thanks @akkomar . I have some thoughts on make some things a bit more "rusty," but we can save that for a subsequent PR.  That would be implementing `From` in a couple places so one can call `into()` to convert between types, a little restructuring to use more `Default::default` calls.

glean_parser/templates/rust_server.jinja2
```diff
@@ -0,0 +1,316 @@
...
+// log type string used to identify logs to process in the Moz Data Pipeline
+const GLEAN_EVENT_MOZLOG_TYPE: &str = "glean-server-event";
+
+// Code below is static, regardless of what is defined in `metrics.yaml`:
+pub struct GleanEventsLogger {
+    // Application Id to identify application per Glean standards
+    pub app_id: String,
+    // Version of application emitting the event
+    pub app_display_version: String,
+    // Channel to differentiate logs from prod/beta/staging/devel
```
badboy, line 24: I'd probably make these proper doc comments (3 slashes):
```suggestion
    /// Application Id to identify application per Glean standards
    pub app_id: String,
    /// Version of application emitting the event
    pub app_display_version: String,
    /// Channel to differentiate logs from prod/beta/staging/devel
```

glean_parser/templates/rust_server.jinja2
```diff
@@ -0,0 +1,316 @@
...
+{% if metrics_by_type['event'] %}
+{# Loop over each event associated with the current ping #}
+{# each event has a specific type and method to create a gleanEvent #}
+{% for event in metrics_by_type['event'] %}
+
+pub struct {{ event|event_type_name }} {
+    // metadata for event in `extra_keys`
+    {% for extra, metadata in event.extra_keys.items() %}
+    {% if extra|event_extra_name == "event" %}
+    pub {{ extra|event_extra_name|snake_case }}: Box<dyn {{ metadata.type|rust_metric_type }}>, // {{ metadata.description|clean_string }}
```
badboy, line 196: Let's make the description a proper doc comment
```suggestion
    /// {{ metadata.description|clean_string }}
    pub {{ extra|event_extra_name|snake_case }}: Box<dyn {{ metadata.type|rust_metric_type }}>,
```

glean_parser/templates/rust_server.jinja2
```diff
@@ -0,0 +1,316 @@
...
+    {% else %}
+    pub {{ extra|event_extra_name|snake_case }}: {{ metadata.type|rust_metric_type }}, // {{ metadata.description|clean_string }}
```
badboy, line 198:
```suggestion
    /// {{ metadata.description|clean_string }}
    pub {{ extra|event_extra_name|snake_case }}: {{ metadata.type|rust_metric_type }},
```

glean_parser/templates/rust_server.jinja2
```diff
@@ -0,0 +1,316 @@
...
+{% endif %}
+{% endfor %}
+{# struct & methods for submitting pings #}
+{% for ping, metrics_by_type in pings.items() %}
+{% if metrics_by_type['event'] %}
+
+{# interface that eligible events for this ping will implement #}
+/// Marker trait for events per ping
+pub trait {{ ping|ping_events_type_name }} {
+    fn glean_event(&self) -> GleanEvent;  // Returns an instance of GleanEvent
```
badboy, line 245: What it returns is given by the type signature

```suggestion
    fn glean_event(&self) -> GleanEvent;
```

glean_parser/templates/rust_server.jinja2
```diff
@@ -0,0 +1,316 @@
...
+pub struct {{ ping|ping_type_name }} {
+    {% for metric_type, metrics in metrics_by_type.items() %}
+    {% if metric_type != 'event' %}
+    {% for metric in metrics %}
+    pub {{ metric|metric_argument_name|snake_case }}: {{ metric_type|rust_metric_type}}, // {{ metric.description|clean_string }}
+    {% endfor %}
+    {% endif %}
+    {% endfor %}
+    {% if metrics_by_type['event'] %}
+    pub event: Option<Box<dyn {{ ping|ping_events_type_name }}>>, // valid event of  `{{ ping|ping_events_type_name }}` for this ping
```
badboy, line 258:
```suggestion
    /// {{ metric.description|clean_string }}
    pub {{ metric|metric_argument_name|snake_case }}: {{ metric_type|rust_metric_type}},
    {% endfor %}
    {% endif %}
    {% endfor %}
    {% if metrics_by_type['event'] %}
    /// valid event of  `{{ ping|ping_events_type_name }}` for this ping
    pub event: Option<Box<dyn {{ ping|ping_events_type_name }}>>,
```

glean_parser/templates/rust_server.jinja2
```diff
@@ -0,0 +1,316 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// AUTOGENERATED BY glean_parser v{{ parser_version }}. DO NOT EDIT.
+
+// Required imports
+use chrono::Utc;
+use serde::{Deserialize, Serialize};
+use serde_json;
+use std::collections::HashMap;
```
badboy, line 12: Couple of places use the full path `std::collections::HashMap`, others use the shortened enabled by this import, `HashMap`.
Feels like we should keep this consistent.

badboy (approved):
generated code looks good to me.
I have a few stylistic nits with respect to the code, but these could also be wrapped into the followup.

glean_parser/templates/rust_server.jinja2
taddes: Good call, not sure why clippy (and I) missed that.

taddes:
Thanks @badboy & @akkomar. I made those doc changes and nits and added some extra documentation, as I did the same on our Sync side. You'll have to re-approve when you have the chance and then feel free to merge, as I am not sure I have write access to the repo. I can then in the near future file another ticket to make some optimizations if we discover some improvements from the Sync side.

PR #774, opened 2024-11-18
Bug 1931277 - Support for Configuring a Subset of Interesting Metrics…

mkmelin:
… to Collect.


travis79 (commented):
This is looking great! Would you mind adding some test coverage for this in `tests/test_parser.py` so we can ensure we don't regress this in the future?

mkmelin:
Updated. FWIW, seems the project is missing some deps. This is what I had to do to make it work:

cd ~/Code/github-glean_parser
mkdir /tmp/mm-env
python3 -m venv /tmp/mm-env
source /tmp/mm-env/bin/activate
pip3 install -r requirements_dev.txt
pip3 install diskcache jsonschema platformdirs
export PYTHONPATH=$PYTHONPATH:`pwd`

pytest tests/test_parser.py

badboy:
those deps are part of the package definition, so a `pip install -e .` should install them.

mkmelin:
> those deps are part of the package definition, so a `pip install -e .` should install them.

Ah yes, that worked!

mkmelin:
Resolved the lint task failure, so this should now be ready to merge.

PR #775, opened 2024-11-19
[go server] enable configurable writer

dmueller:
### problem
The go server implementation is hardcoded to use stdout. As we add more server glean logs, this will make the output have more entries and be harder to read while developing.

### solution
Adjust the go server implementation to allow for a configurable writer. This will enable users of the go server implementation to choose where the logs are written to. When running in production it will have to supply stdout since that is where the ingestion side of things expects to find the logs.

### testing
made similar modifications to the server output locally and confirmed these different scenarios are all valid:
a Writer of `io.Discard` will produce no output
a Writer of a file will write logs to a file
a Writer of `os.Stdout` matches existing behavior

ran `make test-full`

### DIFF
--- a/glean_parser/templates/go_server.jinja2
+++ b/glean_parser/templates/go_server.jinja2
@@ -11,6 +11,7 @@ package glean
 import (
 	"encoding/json"
     "fmt"
+	"io"
 	"strconv"
 	"time"
 
@@ -24,6 +25,7 @@ type GleanEventsLogger struct {
 	AppID             string // Application Id to identify application per Glean standards
 	AppDisplayVersion string // Version of application emitting the event
 	AppChannel        string // Channel to differentiate logs from prod/beta/staging/devel
+	Writer            io.Writer // Writer to output to. Normal operation expects os.Stdout
 }
 
 // exported type for public method parameters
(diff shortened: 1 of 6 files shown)


glean_parser/templates/go_server.jinja2
```diff
@@ -157,7 +159,9 @@ func (g GleanEventsLogger) record(
 	if envelopeErr != nil {
 		panic("Unable to marshal log envelope to json")
 	}
-	fmt.Println(string(envelopeJson))
+	if g.Writer != nil {
+		fmt.Fprintln(g.Writer, string(envelopeJson))
+	}
```
dmueller, line 164: In a separate change, I want to remove the `panic`s from this template and instead bubble up errors. As part of that I will also bubble up an error if `g.Writer == nil`

gleonard-m (approved):

efixler:
Good change +++

I would recommend adding a doc comment somewhere to alert the consumer that in the `File` (or really the `io.Closer`)  case, the consumer will be responsible for closing that handle.

efixler (approved):

badboy (approved):

PR #776, opened 2024-11-20
Bug 1932322 - Support `follows_collection_enabled` for pings

badboy:
Should get some tests.
Also docs.


travis79 (approved):
There's an empty file `tests/test_cli.py`, but other than that this looks good to me.

badboy:
> There's an empty file `tests/test_cli.py`, but other than that this looks good to me.

It's not empty 🙈 (thanks github). It had the executable bit and I figured I remove it while I see it (it is its own commit).

PR #777, opened 2024-11-20
[go server] replace panics with returned errors

dmueller:
### problem
It is idiomatic in go to return errors rather than panic. Panics can cause applications to exit, which is not the desired result if a glean log can't be emitted for any reason.

### solution
Switch the exported functions to return an error. This enables applications that use the glean server implementation to inspect that error return value and take whatever action they deem appropriate.

A couple other things changed in the go server template
* made tabs consistently used in the file
* switched variable declarations to use `:=` syntax

### testing
`make test-full`
also added tests for `Writer = nil` and `Writer = io.Discard` to verify logs are not printed in those scenarios


glean_parser/templates/go_server.jinja2
```diff
@@ -182,28 +195,28 @@ func newGleanEvent(category, name string, extra map[string]string) gleanEvent {
 {% for event in events %}
 
 type {{ event|event_type_name }} struct {
-    {% for extra, metadata in event.extra_keys.items() %}
-    {{ extra|event_extra_name }} {{ metadata.type|go_metric_type }} // {{ metadata.description|clean_string }}
-    {% endfor %}
+	{% for extra, metadata in event.extra_keys.items() %}
```
dmueller, line 198: This template was mixed tabs & spaces, so this whitespace change is to get everything on tabs

glean_parser/templates/go_server.jinja2
```diff
@@ -224,52 +237,52 @@ func (e {{ metric|event_type_name}}) is{{ ping|ping_events_type_name }}() {}
...
 }
 
 // Record and submit `{{ ping }}` ping omitting user request info
 func (g GleanEventsLogger) Record{{ ping|ping_type_name}}WithoutUserInfo(
-    params {{ ping|ping_type_name}},
-) {
-    g.Record{{ ping|ping_type_name }}(defaultRequestInfo, params)
+	params {{ ping|ping_type_name}},
+) error {
+	return g.Record{{ ping|ping_type_name }}(defaultRequestInfo, params)
```
dmueller, line 286: These two generated methods got their signature changed to return error, and the corresponding return statement

tests/test_go_server.py
```diff
@@ -138,8 +138,9 @@ def run_logger(code_dir, code):
     with open(ROOT / "test-go" / "test.go.tmpl", "r") as fp:
         tmpl_code = fp.read()
 
-    tmpl_code = tmpl_code.replace("/* CODE */", code)
+    tmpl_code = tmpl_code.replace("/* CODE */", code).replace("/* IMPORTS */", imports)
 
+    print(tmpl_code)
```
cbguder, line 143: Stray print?

cbguder (approved):

efixler (approved):
nice

chutten (approved):
I'm not a Go expert, but what I understand of these changes looks good to me. (The stray `print` should probably go, though. But it's test-only, so I'm not fussed if it makes it in)

tests/test_go_server.py
dmueller: TY, that was from getting the new tests working

akkomar (commented):
+1 for error handling approach.

I hate to be starting this discussion (lol) but this repo generally uses spaces, not tabs, so these files will have them both mixed eventually. I know Go's recommendation is to use tabs, but IIUC `gofmt` should address that. Is this change useful on your side when working with generated file or is it mostly about consistency across files here?

akkomar:
> Is there a reason to make the jinja templates all use spaces?

Per https://github.com/mozilla/glean_parser/blob/main/.editorconfig#L5-L6 space is default indentation character for these files.

dmueller:
I didn't know about the editorconfig code and am using vscode without a plugin for it, so that's good to know about!

akkomar (approved):

PR #783, opened 2025-01-29
Switch to a pyproject.toml setup, powered by uv

badboy:
This is faster, more coherent, easier to manage and more modern.
~~It comes with one downside though:~~

~~I had to rename the script to `glean-parser` instead of `glean_parser`.~~


chutten (changes requested):
Needs `CHANGELOG`.

What are the implications of the script changing from underscore to hyphen? Would this be considered a breaking change?

badboy:
> Needs `CHANGELOG`.
> 
> What are the implications of the script changing from underscore to hyphen? Would this be considered a breaking change?

None anymore! They fixed the bug.
I probably need to update some docs as well.

.circleci/config.yml
```diff
@@ -14,46 +15,45 @@ commands:
...
+  setup-uv:
     steps:
-      - checkout
       - run:
-          name: environment
+          name: Installing uv
+          command: |
+            UV_SHA256=f6788fd45a25ea1033a671eaa9ece6c9b004ce5c9cf09552effadde567a7d7a5
+            UV=uv-x86_64-unknown-linux-gnu
+            UV_VERSION=0.6.13
```
chutten, line 25: What's the process for upgrading the version of `uv` we use on CI?

README.md
```diff
@@ -16,33 +16,23 @@ code for various integrations, linting and coverage testing.
...
--   jsonschema
--   platformdirs
--   PyYAML
+- Python 3.8 (or later)
 
 ## Usage
 
 ```sh
-$ glean_parser --help
+$ glean-parser --help
```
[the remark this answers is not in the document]
badboy: good catch, forgot to change that back

.circleci/config.yml
badboy: uuh download both, then update here ... yeah, I should document.

chutten (approved):

PR #799, opened 2025-06-24
Bug 1973017 - New lint: HIGHER_DATA_SENSITIVITY_REQUIRED for when an event extra key could potentially contain sensitive data

badboy:


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- New lint: `HIGHER_DATA_SENSITIVITY_REQUIRED` for when an event extra key could potentially contain sensitive data ([bug 1973017](https://bugzilla.mozilla.org/show_bug.cgi?id=1973017))
```
badboy, line 5: happy to bike-shed on the name here!

travis79 (approved):

CHANGELOG.md
chutten: I like it.

glean_parser/lint.py
```diff
@@ -320,6 +320,32 @@ def check_empty_datareview(
         yield "List of data reviews should not contain empty strings or TODO markers."
 
 
+def check_potential_data_sensitivity_required(
+    metric: metrics.Metric, parser_config: Dict[str, Any]
+) -> LintGenerator:
+    # Only looking at event metrics
```
chutten, line 326: If the lint's event-specific, then the name of the fn should say as much.

(( I foresee we'll want something for `object`, too, at some point (if not now) ))

glean_parser/lint.py
```diff
@@ -320,6 +320,32 @@ def check_empty_datareview(
...
+    if not isinstance(metric, metrics.Event):
+        return
+
+    # Not all metrics have `data_sensitivity` defined.
```
chutten, line 330: ref bug 1890648 in the comment?

glean_parser/lint.py
```diff
@@ -320,6 +320,32 @@ def check_empty_datareview(
...
+    if has_data_sensitivity and any(
+        [
+            sensitivity == metrics.DataSensitivity.highly_sensitive
+            for sensitivity in metric.data_sensitivity
+        ]
+    ):
+        return
+
+    # List of potentially sensitive extra key names we want to flag.
+    potential_sensitive_names = ["url", "uri"]
```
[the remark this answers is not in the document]
chutten: Could search for `_id$` and `Id$` and similar, I suppose.

chutten (changes requested):
Looks like the lint won't catch `event` metrics with sensitively-named extras if those metrics don't have any `data_sensitivity` set?

glean_parser/lint.py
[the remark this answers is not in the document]
badboy: That seems very error-prone though. IDs could be anything. Devtools use `session_id`, which is just a random UUID generated every time you open the devtools. That's still technical data and not data-sensitive I'd say and it feels wrong that it would require a `no_lint`.

PR #802, opened 2025-08-12
Bug 1960034 - Object metric: Allow a description in the definition

badboy:
The file is parsed by tests, so without the metrics.py change tests would fail, thus no new test required for this.

### DIFF
--- a/tests/data/object.yaml
+++ b/tests/data/object.yaml
@@ -16,8 +16,10 @@ complex.types:
       - CHANGE-ME@example.com
     expires: never
     structure:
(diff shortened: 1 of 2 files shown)


tests/data/object.yaml
```diff
@@ -16,8 +16,10 @@ complex.types:
       - CHANGE-ME@example.com
     expires: never
     structure:
+      description: An array of numbers
       type: array
       items:
+        description: Every number counts
```
chutten, line 22: -.-

chutten (approved):

PR #821, opened 2025-11-06
DENG-9533 - Go server template: support publishing directly to Pub/Sub

akkomar:
This adds a Go server template for direct-to-Pub/Sub submissions.
For big picture see https://mozilla-hub.atlassian.net/browse/DENG-9533.
This data will be consumed by new Decoder type landed in https://github.com/mozilla/gcp-ingestion/pull/2857.
Next steps after this lands are to test with MARS in stage, iron out wrinkles, and update our documentation.

### DIFF
--- a/glean_parser/go_server.py
+++ b/glean_parser/go_server.py
@@ -21,7 +21,7 @@ Warning: this outputter supports limited set of metrics,
 see `SUPPORTED_METRIC_TYPES` below.
 
 Generated code creates two methods for each ping (`RecordPingX` and `RecordPingXWithoutUserInfo`)
-that are used for submitting events. 
+that are used for submitting events.
 If pings have `event` metrics assigned, they can be passed to these methods.
 """
 
--- a/glean_parser/go_server.py
+++ b/glean_parser/go_server.py
@@ -21,8 +21,8 @@ Warning: this outputter supports limited set of metrics,
 see `SUPPORTED_METRIC_TYPES` below.
 
 Generated code creates two methods for each ping (`RecordPingX` and `RecordPingXWithoutUserInfo`)
-that are used for submitting events. If pings have `event` metrics assigned, they can be
-passed to these methods.
+that are used for submitting events. 
+If pings have `event` metrics assigned, they can be passed to these methods.
 """
 
 from collections import defaultdict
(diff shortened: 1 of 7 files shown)


tests/test_go_server_pubsub.py
akkomar, line 1: This file introduces some duplication with `test_go_server.py`, but I've put it as separate one to be sure that we're not touching current behavior.

glean_parser/templates/go_server.jinja2
```diff
@@ -17,8 +29,10 @@ import (
     "time"
 
     "github.com/google/uuid"
+{% endif %}
 )
 
+{% if transport == "logging" %}
```
badboy: I'm sliiightly irritated by the fact that we sometimes do `if transport == pubsub { } else {}` and sometimes `if transport == logging {} else {}`, especially here where both variants are used in blocks right next to each other.
Should we maybe stick to one way?

glean_parser/templates/go_server.jinja2
```diff
@@ -141,7 +207,9 @@ func (g GleanEventsLogger) createPing(documentType string, config RequestInfo, p
         Payload:           string(payloadJson),
     }, nil
 }
+{% endif %}
```
badboy: We close an `if transport == logging` just to open it in the next line again.

glean_parser/go_server.py
```diff
@@ -11,15 +11,18 @@
...
+- Pub/Sub (go_server_pubsub): Publishes directly to GCP Pub/Sub topics
 
 Warning: this outputter supports limited set of metrics,
 see `SUPPORTED_METRIC_TYPES` below.
 
 Generated code creates two methods for each ping (`RecordPingX` and `RecordPingXWithoutUserInfo`)
-that are used for submitting (logging) them.
-If pings have `event` metrics assigned, they can be passed to these methods.
+that are used for submitting events. If pings have `event` metrics assigned, they can be
+passed to these methods.
```
badboy:
```suggestion
that are used for submitting events. 
If pings have `event` metrics assigned, they can be passed to these methods.
```

and the diff looks much nicer (see also: [Semantic Line Breaks](https://sembr.org/))

badboy (approved):
r+wc. few comments inline, nothing blocking though.

PR #857, opened 2026-08-04
Bug 1908760 - Generate labeled {custom|memory|timing} distribution for mobile 

badboy:
Needed in https://github.com/mozilla/glean/pull/3573

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


travis79 (approved):
A little bit of test coverage in /tests/test_swift.py & test_kotlin.py would be nice for this. No need to re-review as the coverage we currently have in there is pretty simple.

PR #858, opened 2026-08-10
bug 2062403 - Allow 'subcategories' longer than 29 characters

chutten:

### DIFF
--- a/glean_parser/schemas/metrics.2-0-0.schema.yaml
+++ b/glean_parser/schemas/metrics.2-0-0.schema.yaml
@@ -21,7 +21,7 @@ definitions:
 
   dotted_snake_case:
     type: string
-    pattern: "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})*$"
+    pattern: "^[a-z_][a-z0-9_]*(\\.[a-z_][a-z0-9_]*)*$"
     maxLength: 40
 
   event_extra_key:
--- a/tests/data/categories.yaml
+++ b/tests/data/categories.yaml
@@ -1,0 +1,28 @@
+# Any copyright is dedicated to the Public Domain.
+# https://creativecommons.org/publicdomain/zero/1.0/
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+
+we_used_to_have_subcategory_length:
+  but_we_dont_any_more: &defaults
+    type: counter
+    expires: never
+    description: A test metric
+    bugs: [https://bugzil.la/2062403]
+    data_reviews: [https://www.example.com]
+    notification_emails: [glean-team@mozilla.com]
+
+it.may_matter_that_theres_a_dot_here:
+  so_we_will_test_that_too:
+    <<: *defaults
+
+# Subcategory length 30, Category length 40
+start.abcdefghijklmnopqrstuvwxyzabcd.end:
+  no_problem:
+    <<: *defaults
+
+a.b:
+  really_short_categories_are_allowed:
+    <<: *defaults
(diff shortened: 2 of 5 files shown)


glean_parser/schemas/metrics.2-0-0.schema.yaml
```diff
@@ -21,7 +21,7 @@ definitions:
 
   dotted_snake_case:
     type: string
-    pattern: "^[a-z_][a-z0-9_]{0,29}(\\.[a-z_][a-z0-9_]{0,29})*$"
+    pattern: "^[a-z_][a-z0-9_]{1,40}(\\.[a-z_][a-z0-9_]{1,39})*$"
```
badboy: Do we even need the upper limits here? It's correct I guess, but maybe overly complicated.
The total length is limited due to `maxLength`.
We want to express that: 
* Needs to start with a lower-case character (or `_` for some reason!)
* A dot is followed by an lower-case alphanumeric character (or `_` again!)

Your changed regex btw disallows a previously alowed category of `a.b`

tests/data/categories.yaml
```diff
@@ -0,0 +1,14 @@
+# Any copyright is dedicated to the Public Domain.
+# https://creativecommons.org/publicdomain/zero/1.0/
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+we_used_to_have_subcategory_length:
```
badboy, line 8: This doesn't have any "subcategory", by which I think you mean dot-separated parts.

Maybe additionally this file should list the edge-cases I mentioned above (`a.b`), an actual "subcategory" longer than 29 characters, ...

badboy (changes requested):

tests/data/categories.yaml
badboy: Ah, I was somewhat wrong here. Of course this acts as _a single_ "subcategory" (hey, we never defined "subcategory"!) -- which now is longer. Still, I feel like we should cover the other cases.

tests/data/categories.yaml
```diff
@@ -0,0 +1,14 @@
...
+  but_we_dont_any_more:
+    type: counter
+    expires: never
+    description: A test metric
+    bugs: [https://bugzil.la/2062403]
+    data_reviews: [https://www.example.com]
+    notification_emails: [glean-team@mozilla.com]
```
badboy, line 15:
```suggestion
    notification_emails: [glean-team@mozilla.com]

# Total length: 39
# Subcategory: 29
start.abcdefghijklmnopqrstuvwxyzabc.end:
  but_we_dont_any_more: &defaults
    type: counter
    expires: never
    description: A test metric
    bugs: [https://bugzil.la/2062403]
    data_reviews: [https://www.example.com]
    notification_emails: [glean-team@mozilla.com]

# Total length: 40
# Subcategory: 30
start.abcdefghijklmnopqrstuvwxyzabcd.end:
  but_we_dont_any_more:
    <<: *defaults

a.b:
  but_we_dont_any_more:
    <<: *defaults
```

glean_parser/schemas/metrics.2-0-0.schema.yaml
chutten: Good points. Whether or not `a.b` is a good category name doesn't allow me to escape that this is a non-backwards-compatible change. Lemme move these groups to `*` so subcategories must all have at least one character (and, if so, it's not a digit) and the total length is only enforced by `maxLength`

tests/data/categories.yaml
chutten: Fair 'nuff, this only exercises the first regex branch. I'll add an extra couple cases.

badboy (approved):

